PR #283, opened 2018-08-16
James/set settings earlier

JamesMBallard:
Assigned to @ldanzinger.

@lsmallwood and @ldanzinger to review.

Summary of changes:
 * Live location is now the default setting. If no setting is available in `DsaAppConfig.json`, live location is the default option - `"SimulateLocation": "false"`
 * For both settings (live/simulated), the position source is now only created once. Previously we would create a position source and then recreate it right away if the user requested live or simulated and our first choice was wrong.
 * Refactor work in DsaController. Each tool's settings are set right after the tool is added to the tool manager. This allows us to get app settings applied earlier. This allows us to not have to create a position source once as a first guess and then recreate it immediately based on the settings.
 * Minor refactoring work in `MessageFeedsController` since assigning the tool properties sooner caused some problems since we don't have the geoView set at that point. We now defer that work to when the geoView is available.
 * Requires some updates to the toolkit as well - PR coming.


Shared/DsaController.cpp
```diff
@@ -47,6 +47,7 @@
 #include <QSettings>
 
 using namespace Esri::ArcGISRuntime;
+using Esri::ArcGISRuntime::Toolkit::ToolManager;
```
lsmallwood: maybe just do `using Esri::ArcGISRuntime::Toolkit;`?

Shared/LayerCacheManager.h
```diff
@@ -70,8 +70,8 @@ private slots:
   QJsonArray m_layers;
   QJsonArray m_inputLayerJsonArray;
   bool m_initialLoadCompleted = false;
-  AddLocalDataController* m_localDataController;
-  Esri::ArcGISRuntime::Scene* m_scene;
+  AddLocalDataController* m_localDataController = nullptr;
+  Esri::ArcGISRuntime::Scene* m_scene  = nullptr;
```
lsmallwood: two spaces before the `=` here

Shared/messages/MessageFeedsController.cpp
```diff
@@ -87,6 +87,10 @@ MessageFeedsController::~MessageFeedsController()
 void MessageFeedsController::setGeoView(GeoView* geoView)
 {
   m_geoView = geoView;
+
+  // now that the geoview is ready, setup the feeds
+  if (m_geoView && m_messageFeeds->rowCount() == 0)
```
lsmallwood: I guess it's possible that there will not be any feeds. But presumably this only gets called once, when the geoview is set up, so we shouldn't get repeated calls to `setupFeeds();`?

Shared/messages/MessageFeedsController.h
```diff
@@ -96,6 +98,7 @@ class MessageFeedsController : public Esri::ArcGISRuntime::Toolkit::AbstractTool
   QList<DataListener*> m_dataListeners;
   QString m_resourcePath;
   LocationBroadcast* m_locationBroadcast = nullptr;
+  QVariantList m_messageFeedProperties;
```
lsmallwood: Can't this be a `QStringList`?

Shared/DsaController.cpp
JamesMBallard: I tried it and it doesn't work, it says using must be used with types, not a namespace. I can do this, how's this?

```cpp
using namespace Esri::ArcGISRuntime::Toolkit;
```

Shared/messages/MessageFeedsController.cpp
JamesMBallard: Correct as far as I know. `m_messageFeeds->rowCount() == 0` was the original check and I've just expanded it to include the geoview check.

Shared/DsaController.cpp
lsmallwood: seems better

Shared/messages/MessageFeedsController.h
JamesMBallard: It can but that would require more refactoring. The existing code processes it like this. Do you want me to make that change?

PR #284, opened 2018-08-20
James/360 viewshed

JamesMBallard:
Assigned to @michael-tims. Please review. Also @lsmallwood to review if you have time.

This updates the DSA app to use the Runtime's new support for 360 degree viewsheds at the 100.3 release. The UI slider now allows up to 359 degrees when not in 360 mode.


JamesMBallard:
@lsmallwood I actually hadn't looked there. I had a quick look and it doesn't look like anything can be cleaned up there, it didn't know anything about the multiple viewsheds that were being created internally.

PR #287, opened 2018-09-12
DSA:

JamesMBallard:
Mitigate the GeoElement base class changes.
Provide helpers and a wrapper class that provide the
now-missing signal on GeoElement.

@michael-tims @lsmallwood @ldanzinger please review. I've added a new GeoElementUtils set of helpers and a wrapper class to minimize the code changes needed in the DSA app. Mostly what we need to do and can't do anymore is set the parent and connect directly to a GeoElement's signals.

Blocked for now, requires changes not in the Runtime SDK yet.

### DIFF
--- a/Shared/GeometryQuadtree.cpp
+++ b/Shared/GeometryQuadtree.cpp
@@ -1,3 +1,4 @@
+
 /*******************************************************************************
  *  Copyright 2012-2018 Esri
  *
@@ -18,6 +19,7 @@
 #include "pch.hpp"
 
 #include "GeometryQuadtree.h"
+#include "GeoElementUtils.h"
 
 // C++ API headers
 #include "Envelope.h"
@@ -100,7 +102,6 @@ GeometryQuadtree::GeometryQuadtree(const Envelope& extent,
  */
 GeometryQuadtree::~GeometryQuadtree()
 {
-
 }
 
 /*!
@@ -150,9 +151,9 @@ QList<Geometry> GeometryQuadtree::candidateIntersections(const Envelope& extent)
     auto findIt = m_elementStorage.find(id);
     if (findIt != m_elementStorage.end())
     {
-      GeoElement* element = findIt.value();
+      GeoElementSignaler* element = findIt.value();
       if (element)
-        results.push_back(element->geometry());
+        results.push_back(element->m_geoElement->geometry());
     }
   }
 
@@ -181,9 +182,9 @@ QList<Geometry> GeometryQuadtree::candidateIntersections(const Point& location) 
     auto findIt = m_elementStorage.find(id);
     if (findIt != m_elementStorage.end())
     {
-      GeoElement* element = findIt.value();
+      GeoElementSignaler* element = findIt.value();
       if (element)
-        results.push_back(element->geometry());
+        results.push_back(element->m_geoElement->geometry());
     }
   }
 
@@ -206,11 +207,11 @@ void GeometryQuadtree::buildTree(const Envelope& extent)
   auto itEnd = m_elementStorage.cend();
   for (; it != itEnd; ++it)
   {
-    GeoElement* element = it.value();
+    GeoElementSignaler* element = it.value();
     if (!element)
       continue;
 
-    const Geometry wgs84 = GeometryEngine::project(element->geometry(), SpatialReference::wgs84());
+    const Geometry wgs84 = GeometryEngine::project(element->m_geoElement->geometry(), SpatialReference::wgs84());
     m_tree->assign(wgs84.extent(), it.key(), m_maxLevels);
   }
 
@@ -225,11 +226,11 @@ void GeometryQuadtree::buildTree(const Envelope& extent)
  */
 void GeometryQuadtree::handleGeometryChange(int changedId)
 {
-  const GeoElement* changedElement = m_elementStorage.value(changedId);
+  const GeoElementSignaler* changedElement = m_elementStorage.value(changedId);
   if (!changedElement)
     return;
 
-  const Geometry wgs84Geom = GeometryEngine::project(changedElement->geometry(), SpatialReference::wgs84());
+  const Geometry wgs84Geom = GeometryEngine::project(changedElement->m_geoElement->geometry(), SpatialReference::wgs84());
   const Envelope wgs84Extent = wgs84Geom.extent();
 
   // if the extent of the changed geom is the same or smaller than the existing tree, it can still be used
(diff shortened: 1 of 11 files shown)


Shared/analysis/GeoElementViewshed360.cpp
```diff
@@ -62,7 +64,7 @@ constexpr double c_defaultOffsetZ = 5.0;
 GeoElementViewshed360::GeoElementViewshed360(GeoElement* geoElement, AnalysisOverlay* analysisOverlay,
                                              const QString& headingAttribute, const QString& pitchAttribute, QObject* parent) :
   Viewshed360(new GeoElementViewshed(geoElement, c_defaultHorizontalAngle, c_defaultVerticalAngle, c_defaultMinDistance, c_defaultMaxDistance, 0.0, 0.0, parent), analysisOverlay, parent),
-  m_geoElement(geoElement),
+  m_geoElementSignaler(new GeoElementSignaler(geoElement, GeoElementUtils::toQObject(geoElement))),
```
JamesMBallard: This is a self-destructing helper (m_geoElementSignaler is a `QPointer`). It should delete/invalidate itself whenever the GeoElement goes away.

Shared/utilities/GeoElementUtils.h
```diff
@@ -0,0 +1,55 @@
...
+
+class GeoElementSignaler : public QObject
+{
+  Q_OBJECT
+
+public:
+  GeoElementSignaler(Esri::ArcGISRuntime::GeoElement* geoElement, QObject* parent = nullptr);
+  ~GeoElementSignaler();
+
+  Esri::ArcGISRuntime::GeoElement* m_geoElement = nullptr;
```
lsmallwood: we should probably encapsulate this member with a getter and setter (if required)

Shared/utilities/GeoElementUtils.cpp
```diff
@@ -0,0 +1,135 @@
...
+ ******************************************************************************/
+
+// PCH header
+#include "pch.hpp"
+
+#include "GeoElementUtils.h"
+
+// C++ API headers
+#include "GeoElement.h"
+#include "Feature.h"
```
lsmallwood: alphabetic order

Shared/utilities/GeoElementUtils.cpp
```diff
@@ -0,0 +1,135 @@
...
+  else if (dynamic_cast<EncFeature*>(m_geoElement))
+  {
+    connect(static_cast<EncFeature*>(m_geoElement), &EncFeature::geometryChanged,
+            this, &GeoElementSignaler::geometryChanged);
+  }
+  else if (dynamic_cast<WmsFeature*>(m_geoElement))
+  {
+    connect(static_cast<WmsFeature*>(m_geoElement), &WmsFeature::geometryChanged,
+            this, &GeoElementSignaler::geometryChanged);
+  }
```
lsmallwood: I think we should add an else for the error case here - e.g. if we add a new type of geo element

Shared/utilities/GeoElementUtils.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+  for (auto* geoElement : geoElements)
+  {
+    toQObject(geoElement)->setParent(parent);
+  }
+}
+
+/*!
+  \fn void Dsa::GeoElementUtils::setParent(Esri::ArcGISRuntime::GeoElement* geoElement, QObject* parent)
+  \brief Allows the \a parent to be applied to \a geoElement.
```
lsmallwood: "to the \a geoElement."

Shared/utilities/GeoElementUtils.cpp
```diff
@@ -0,0 +1,135 @@
...
+  }
+}
+
+/*!
+  \fn QObject* Dsa::GeoElementUtils::toQObject(Esri::ArcGISRuntime::GeoElement* geoElement)
+  \brief Returns a QObject* from a GeoElement.
+ */
+QObject* GeoElementUtils::toQObject(Esri::ArcGISRuntime::GeoElement* geoElement)
+{
+  if (dynamic_cast<Feature*>(geoElement))
```
lsmallwood: I think you raised an issue for a "geo element type enum" - looks like it would make this a lot cleaner

Shared/GeometryQuadtree.cpp
```diff
@@ -150,9 +151,9 @@ QList<Geometry> GeometryQuadtree::candidateIntersections(const Envelope& extent)
     auto findIt = m_elementStorage.find(id);
     if (findIt != m_elementStorage.end())
     {
-      GeoElement* element = findIt.value();
+      GeoElementSignaler* element = findIt.value();
       if (element)
-        results.push_back(element->geometry());
+        results.push_back(element->m_geoElement->geometry());
```
michael-tims: @JamesMBallard In this case, is the `GeoElementSignaler` required or can the code here remain untouched?  There are others in this PR that are similar.

JamesMBallard: I need to persist the object with the signal so we can listen for it. Unfortunately that means we need change the member to this. Not too bad and the `GeoElementSignaler` just wraps the geoElement.

```cpp
QHash<int, GeoElementSignaler*> m_elementStorage;
```

Shared/utilities/GeoElementUtils.h
JamesMBallard: @lsmallwood I avoided that to keep this as simple a wrapper as possible. It really only exists for the signal and nothing more. I'll add it to maintain proper coding practices.

JamesMBallard:
@lsmallwood @michael-tims I pushed up some code review fixes.

Shared/GeometryQuadtree.cpp
michael-tims: @JamesMBallard But it looks like for this particular code that `GeoElement` wasn't being persisted before these changes.  It looks like we're just pulling out the current geometry and hooking up to the `geometryChanged` signal.  I assume that's being done elsewhere.  Let's chat offline if I'm missing something obvious.

JamesMBallard: @michael-tims the GeoElement was being persisted before in the `m_elementStorage` member. This local `element` is being pulled out of `m_elementStorage`.

michael-tims:
@JamesMBallard Ah I see now, the element cache now contains `GeoElementSignaler` objects instead of `GeoElement`.  Makes sense.

Shared/utilities/GeoElementUtils.cpp
```diff
@@ -0,0 +1,151 @@
...
+GeoElement* GeoElementSignaler::geoElement() const
+{
+  return m_geoElement;
+}
+
+/*!
+  \fn void Dsa::GeoElementUtils::setParent(const QList<Esri::ArcGISRuntime::GeoElement*> geoElements, QObject* parent)
+  \brief Allows the \a parent to be applied to all the \a geoElements.
+ */
+void GeoElementUtils::setParent(const QList<Esri::ArcGISRuntime::GeoElement*> geoElements, QObject* parent)
```
michael-tims: `const QList<Esri::ArcGISRuntime::GeoElement*> geoElements` -> `const QList<Esri::ArcGISRuntime::GeoElement*>& geoElements`

JamesMBallard: Good catch @michael-tims. Fixed.

PR #294, opened 2018-10-18
initial commit adding KML support to DSA

ldanzinger:
@JamesMBallard please review/merge. Adding KML into DSA. This exposes it to the add local data tool, caches it via the layer cache manager, and displays in the layer list control. Everything seems to "just work", which is nice


Shared/AddLocalDataController.cpp
```diff
@@ -68,7 +70,7 @@ namespace Dsa
 const QString AddLocalDataController::LOCAL_DATAPATHS_PROPERTYNAME = "LocalDataPaths";
 const QString AddLocalDataController::DEFAULT_ELEVATION_PROPERTYNAME = "DefaultElevationSource";
 
-const QString AddLocalDataController::s_allData = QStringLiteral("All Data (*.geodatabase *.tpk *.shp *.gpkg *.mmpk *.slpk *.vtpk *.img *.tif *.tiff *.i1, *.dt0 *.dt1 *.dt2 *.tc2 *.geotiff *.hr1 *.jpg *.jpeg *.jp2 *.ntf *.png *.i21 *.ovr *.markup *.sid)");
+const QString AddLocalDataController::s_allData = QStringLiteral("All Data (*.geodatabase *.tpk *.shp *.gpkg *.slpk *.img *.tif *.tiff *.i1, *.dt0 *.dt1 *.dt2 *.tc2 *.geotiff *.hr1 *.jpg *.jpeg *.jp2 *.ntf *.png *.i21 *.ovr *.markup *.sid *.kml *.kmz)");
```
JamesMBallard: Please confirm you meant to drop some file types here.

Shared/AddLocalDataController.cpp
```diff
@@ -817,6 +827,40 @@ void AddLocalDataController::createTiledLayer(const QString& path, int layerInde
...
+  connect(kmlLayer, &KmlLayer::errorOccurred, this, &AddLocalDataController::errorOccurred);
+
+  if (autoAdd)
+  {
+    auto operationalLayers = Toolkit::ToolResourceProvider::instance()->operationalLayers();
+    if (operationalLayers)
+      operationalLayers->append(kmlLayer);
+
+    emit layerSelected(kmlLayer);
+    Q_UNUSED(layerIndex)
```
JamesMBallard: This is not needed. The arg is used in this method, just not this block.

Shared/DsaResources.cpp
```diff
@@ -292,6 +292,14 @@ QUrl DsaResources::iconHome() const
   return QStringLiteral("qrc:/Resources/icons/xhdpi/ic_menu_home_dark.png");
 }
 
+/*!
+  \brief Returns the URL to the "KML" icon.
+ */
+QUrl DsaResources::iconKml() const
+{
+  return QStringLiteral("qrc:/Resources/icons/xhdpi/kml.png");
```
JamesMBallard: This should return a QUrl to avoid an implicit cast.

Perhaps this?
```cpp
  return QUrl(QStringLiteral("qrc:/Resources/icons/xhdpi/kml.png"));
```

Shared/LayerCacheManager.cpp
```diff
@@ -291,6 +292,11 @@ void LayerCacheManager::layerToJson(Layer* layer)
   if (markupLayer)
     layerPath = markupLayer->path();
 
+  // Get KmlLayers
+  auto kmlLayer = dynamic_cast<KmlLayer*>(layer);
+  if (kmlLayer)
+    layerPath = kmlLayer->dataset()->url().toLocalFile();
```
JamesMBallard: For now this will always be a local file, but is it worth checking to make sure the url is a local file first? That check would help in the future if we support online datasets, but may be overkill for now.

Shared/AddLocalDataController.cpp
ldanzinger: yes i did

ldanzinger:
@JamesMBallard thanks. review comments should be addressed

Shared/LayerCacheManager.cpp
```diff
@@ -291,6 +292,15 @@ void LayerCacheManager::layerToJson(Layer* layer)
   if (markupLayer)
     layerPath = markupLayer->path();
 
+  // Get KmlLayers
+  auto kmlLayer = dynamic_cast<KmlLayer*>(layer);
+  if (kmlLayer)
+  {
+    const QUrl kmlUrl = kmlLayer->dataset()->url();
+    const QString localKmlFile = kmlUrl.toLocalFile();
+    layerPath = localKmlFile.isEmpty() ? kmlUrl.toString() : localKmlFile;
```
JamesMBallard: There's a QUrl [isLocalFile](https://doc-snapshots.qt.io/qt5-5.11/qurl.html#isLocalFile) that would work better here. What do you think?

Something like this
```cpp
const QUrl kmlUrl = kmlLayer->dataset()->url();
layerPath = kmlUrl.isLocalFile() ? kmlUrl.toLocalFile() : kmlUrl.toString();
```

Shared/AddLocalDataController.cpp
```diff
@@ -808,15 +809,47 @@ void AddLocalDataController::createTiledLayer(const QString& path, int layerInde
...
+   \li \a path - The path to the local data source.
+   \li \a layerIndex - The index for which the layer will be added to the operational layer list.
+   \li \a visible - Whether the layer should be visible by default.
+   \li \a autoAdd - Whether the layer will be automatically added to the operational layer list.
+        If \c false, it will not add automatically. Instead, a signal will emit once the Layer has
+        been constructed.
+ \endlist
+*/void AddLocalDataController::createKmlLayer(const QString& path, int layerIndex, bool visible, bool autoAdd)
+{
+  KmlDataset* kmlDataset = new KmlDataset(path, this);
```
JamesMBallard: I missed this on the first pass, but this should use the proper conversion for QUrl and local files.

```suggestion
  KmlDataset* kmlDataset = new KmlDataset(QUrl::fromLocalFile(path), this);
```

ldanzinger:
good idea @JamesMBallard. updated

PR #308, opened 2019-06-21
Ldanzinger/mspk updates

ldanzinger:
@JamesMBallard / @lsmallwood please review

Summary of changes:
- Updating LayerCacheManager so that it works with a scene from MSPK or from default programatic scene
- Moving Open Scene tool to the "Map" toolbar instead of the main toolbar
- Adding a Scene Layer icon since it defaulted to unknown and displayed nothing
- Added option to create a "default scene" from within the Open Scene tool. This is needed because if you open a Package and are viewing a scene, but then you want to just go back to not using a MSPK any longer and just adding data ad hoc, there is no way to get back other than manually editing the config and restarting. This will allow the user to press a button and go back to the default scene.
- removed explicit width of the alert/report tool to fix centering issue


Shared/DsaController.h
```diff
@@ -17,6 +17,12 @@
 #ifndef DSACONTROLLER_H
 #define DSACONTROLLER_H
 
+// toolkit headers
+#include "AbstractTool.h"
```
lsmallwood: You probably don't need the `QObject` include anymore with this

Shared/packages/OpenMobileScenePackageController.cpp
```diff
@@ -253,6 +263,48 @@ void OpenMobileScenePackageController::unpack()
...
 
+/*!
+  \brief Creates a new scene with the default basemap and default surface.
+ */
+void OpenMobileScenePackageController::createDefaultScene()
+{
+  // obtain other required tools
+  auto basemapTool = Toolkit::ToolManager::instance().tool<BasemapPickerController>();
+  auto layerCacheTool = Toolkit::ToolManager::instance().tool<LayerCacheManager>();
+  auto dsaController = Toolkit::ToolManager::instance().tool<DsaController>();
```
lsmallwood: I don't really like adding dependencies on all of these tools. But maybe can't be avoided

ldanzinger: Will try and shuffle this around to DsaController instead

Shared/DsaController.cpp
```diff
@@ -307,28 +306,72 @@ bool DsaController::setInitialLocationFromConfig()
...
+  // add elevation
+  m_cacheManager->addElevation(m_dsaSettings);
+
+  // remove any layers if automatically added
+  if (!newScene->operationalLayers()->isEmpty())
+  {
+    for (auto layer : *newScene->operationalLayers())
+    {
+      delete layer;
+      layer = nullptr;
```
JamesMBallard: This is not needed. You're clearing the collection a few lines after this.

Shared/DsaController.h
```diff
@@ -49,6 +52,9 @@ class DsaController : public QObject
 
   void init(Esri::ArcGISRuntime::GeoView* geoView);
 
+  Esri::ArcGISRuntime::Viewpoint defaultViewpoint();
```
JamesMBallard: You can probably forward-declare this class and avoid the include here.

Shared/LayerCacheManager.cpp
```diff
@@ -81,39 +82,26 @@ LayerCacheManager::LayerCacheManager(QObject* parent) :
...
-      emit jsonToLayerCompleted(layer);
+    m_scene = Toolkit::ToolResourceProvider::instance()->scene();
+    m_layers = QJsonArray();
+    m_inputLayerJsonArray = QJsonArray();
+    m_initialLayerCache.clear();
+    connectSignals();
+
+    // only add initial layers on initial load. Once the user selects a new scene, the layer list will be cleared
+    auto mobileSceneTool = Toolkit::ToolManager::instance().tool<OpenMobileScenePackageController>();
+    if (mobileSceneTool->userSelected())
```
JamesMBallard: Is there any chance `mobileSceneTool` could be null?

Shared/LayerCacheManager.cpp
```diff
@@ -365,6 +319,117 @@ void LayerCacheManager::addExcludedPath(const QString& exludedPath)
...
+{
+  const QVariant elevationData = properties.value(ELEVATION_PROPERTYNAME);
+  const QStringList pathList = elevationData.toStringList();
+
+  // If size is 1, it could be a TPK or raster
+  if (pathList.length() == 1)
+  {
+    // Get the string
+    const QString elevationSource = pathList.at(0);
+    QFileInfo elevationSourceInfo(elevationSource);
```
JamesMBallard: can this be const?

Shared/LayerCacheManager.cpp
```diff
@@ -365,6 +319,117 @@ void LayerCacheManager::addExcludedPath(const QString& exludedPath)
...
+
+    // connect signals
+    m_dataChangedConnection = connect(m_scene->operationalLayers(), &LayerListModel::dataChanged, this, &LayerCacheManager::onLayerListChanged); // layer objects have been added or changed
+    m_layerAddedConnection = connect(m_scene->operationalLayers(), &LayerListModel::layerAdded, this, &LayerCacheManager::onLayerListChanged); // layer objects have been added
+    m_layerRemovedConnection = connect(m_scene->operationalLayers(), &LayerListModel::layerRemoved, this, &LayerCacheManager::onLayerListChanged); // layer has been removed
+    m_layoutChangedConnection = connect(m_scene->operationalLayers(), &LayerListModel::layoutChanged, this, &LayerCacheManager::onLayerListChanged); // order changed
+    m_modelResetConnection = connect(m_scene->operationalLayers(), &LayerListModel::modelReset, this, &LayerCacheManager::onLayerListChanged); // order changed
+  }
+
+  if (m_localDataController)
```
JamesMBallard: This is optional but if you consider inverting this check, along with `if (m_scene)` above, you can reduce the nesting.

For example
```c++
if (m_scene)
{
  // lots of logic
  if (m_localDataController)
  {
    // lots of logic
  }
}
```
to
```c++
if (!m_scene)
  return;

// lots of logic

if (!m_localDataController)
  return;

// lots of logic
```

Shared/packages/OpenMobileScenePackageController.h
```diff
@@ -74,6 +74,8 @@ class OpenMobileScenePackageController : public Esri::ArcGISRuntime::Toolkit::Ab
 
   QAbstractListModel* packages() const;
 
+  bool userSelected() const { return m_userSelected; }
```
JamesMBallard: Move the implementation to the .cpp file. The compiler may inline it anyway.

Shared/DsaController.cpp
JamesMBallard: To clarify, `delete layer;` makes sense, `layer = nullptr;` is not needed.

Shared/LayerCacheManager.cpp
ldanzinger: yes, perhaps. it's instantiated in a qml file so we can't exactly control order. i'll add a check

ldanzinger:
thanks guys. updated. I addressed James's comments and moved the dependencies out of the OpenMobileScenePackage tool. I agree it is cleaner

PR #299, opened 2018-12-06
Lsmallwood/mspk support

lsmallwood:
@JamesMBallard, @ldanzinger - please review the changes to support `MobileScenePackage`

PR #320, opened 2020-01-21
James/100.7 launch

anmacdonald:
I was receiving errors on duplicate state names. It looks as if the state `basemapIcon.toolName` was referenced twice in this list. Could you verify I have removed the correct one?


JamesMBallard:
@anmacdonald confirmed this is correct. Looks like this was introduced back [here](https://github.com/Esri/dynamic-situational-awareness-qt/pull/299/files), specifically in https://github.com/Esri/dynamic-situational-awareness-qt/commit/b6381a0e815b10f7e0fd873a4c079bc2bccedbe0.

PR #342, opened 2020-06-09
Anmacdonald/regression fix

anmacdonald:
Reverting back to ModelSceneSymbol. 

Note that the README.md changes are just along for the ride, looks like they're not in v.next for some reason.


anmacdonald:
Tested with Android 29. `QDir::tempPath` behaves. All good.

PR #344, opened 2020-07-30
Updated to SDK 100.9 and Qt 5.15. Made changes to support both.

anmacdonald:
Updgraded to 100.9 and Runtime 100.9


Vehicle/Android/Android.pri
```diff
@@ -24,4 +24,7 @@ OTHER_FILES +=     $$ANDROID_PACKAGE_SOURCE_DIR/res/drawable-ldpi/icon.png     $
 DISTFILES += \
     $$PWD/AndroidManifest.xml \
     $$PWD/build.gradle \
-    $$PWD/res/values/libs.xml \
+    $$PWD/gradle/wrapper/gradle-wrapper.jar \
+    $$PWD/gradle/wrapper/gradle-wrapper.properties \
+    $$PWD/gradlew \
+    $$PWD/gradlew.bat
```
JamesMBallard, line 30: Do we need all these files? Are they auto-generated as part of the build? We included them in our [AR samples](https://github.com/Esri/arcgis-runtime-samples-qt/tree/master/ArcGISRuntimeSDKQt_CppSamples/AR/DisplayScenesInTabletopAR/android), but I want to be sure they are required.

anmacdonald: These were created as part of the "Android Template Wizard" and not the build step. Qt-creator added these files to `DIST_FILES` automatically, so I've kept them as part of the build as that seemed to be Qt's intention.

PR #345, opened 2020-07-31
Improvements to app documentation

mikewilburn:
This PR:
- renames `docs/index.md` to `docs/README.md`
- adds table of contents to (2) README files. The TOC was generated using MDTOC.
- links to `docs/README.md` from the root README.md
- makes a number of formatting corrections to root README.md (924719d & ef1ac8b)
- describes patch changes in RELEASE.md
- increments app version to the applicable patch number

@anmacdonald I might need your help with incrementing the app version in various locations. I'm not positive I caught them all in be155c3.


mikewilburn:
@esreli the TOC in the detailed doc was a bit tricky. Since the doc it makes use of multiple h1s, it seemed useful to leave the generation defaulted to `firsth1:1`. I then placed the TOC at the immediate top of the document titled by "Contents". See what you think.

docs/README.md
```diff
@@ -1,3 +1,46 @@
...
+      - [![](./images/dsa-icon-viewshed-32.png) Viewshed](#imagesdsa-icon-viewshed-32png-viewshed)   
+      - [![](./images/dsa-icon-line-of-site-32.png) Line of sight](#imagesdsa-icon-line-of-site-32png-line-of-sight)   
+   - [Alerts and conditions](#alerts-and-conditions)   
+      - [New alert notification](#new-alert-notification)   
+      - [![](./images/dsa-icon-view-alerts-32.png) Alerts view](#imagesdsa-icon-view-alerts-32png-alerts-view)   
+      - [![](./images/dsa-icon-alert-conditions-32.png) Conditions](#imagesdsa-icon-alert-conditions-32png-conditions)   
+   - [Collaboration](#collaboration)   
+      - [![](./images/dsa-icon-create-reports-32.png) Create report tools](#imagesdsa-icon-create-reports-32png-create-report-tools)   
+      - [![](./images/dsa-icon-observation-32.png) Observation report](#imagesdsa-icon-observation-32png-observation-report)   
+      - [![](./images/dsa-icon-markup-tools-32.png) Markup tools](#imagesdsa-icon-markup-tools-32png-markup-tools)
```
esreli, line 34: There are some headers in the document that contain images. These images break the hyperlink from the TOC to the section header. Can we figure out a different solution here? Perhaps the image is moved to the first line of the section?

anmacdonald:
@mikewilburn you've discovered all the files that needed updated! The only missing piece of the puzzle is that `android:versionCode`  needs updated with `andoid:versionName`. Both are documented [here](https://developer.android.com/guide/topics/manifest/manifest-element). It's an annoying gotcha, I've pushed the change up to your branch.


FYI, `v.next` is currently at `v1.2.0`, if you want you could branch off of that and then merge back into `v.next` as well. Your changes could become part of the upcoming `1.2.0` release and you wouldn't have to worry about a `1.1.3`.

docs/README.md
anmacdonald: Looks like the link is:
#-add-data


But the TOC link generated is:
#imagesdsa-icon-add-data-32png-add-data

Unfortuante the TOC generator doesn't just work out the box!

mikewilburn:
Thanks for pushing up the Android version number changes @anmacdonald!

Regarding the broken TOC links, in my latest commit, I've taken @esreli's suggestion of removing the icon images from the heading lines and incorporating them immediately below into the content's body. This allows us rely on automatic generation of the TOC rather than having to manually override it each time. See what you think.

mikewilburn:
> FYI, v.next is currently at v1.2.0, if you want you could branch off of that and then merge back into v.next as well. Your changes could become part of the upcoming 1.2.0 release and you wouldn't have to worry about a 1.1.3.

@anmacdonald if it's no different to you, I can simply rebase `v.next` with these changes after this PR's been merged. That might be a tad easier at this point.

PR #346, opened 2020-09-14
Anmacdonald/missing mspk

anmacdonald:
In some instances, paths belonging to SLPKs are serlialized to the layers output when they should be skipped. Unfortunately by the time we reach this area of functionality, the source of the layer has been lost and the path is relative to the root of the SLPK, not the position on the filesystem.

This workaround is to prevent serialization of files that are missing. Take the reasonable precaution of checking if a path _could_ be represented locally, and if that file actually exists. Under the assumption not all paths may be local, but unclear on that.


lsmallwood (approved):
LGTM

PR #347, opened 2020-09-17
Fixed issue in exlclusion folder code.

anmacdonald:
All layers are excluded in cases where the exclusion folder list contains an empty string. 
An empty string is inserted into the exclusion folder list when there is no `PackageDirectory` key on the config.


lsmallwood (approved):
LGTM

PR #349, opened 2020-09-22
DSA 1.1.4

anmacdonald:
DSA 1.1.4 includes: 

* Minumum Runtime Qt SDK version is 100.9.
* Minimum Qt SDK version is Qt 5.15.0.
* Minimum Android SDK version is now 21, target SDK version is 29.
* Fixed erronious "missing file" dialog appearing on startup when opening an MSPK. This was a layer-to-json issue when saving the config.
* Reverted the "ModelSceneSymbol to PictureMarkerSymbol" change. A camera bug was was introduced because a rotation-expression can not be used in place of a heading-expression when working with an OrbitGeoElementCameraController. We will need to park this upgrade until an enhancement is exposed by ArcGISRuntime.
* Fixed bug where layers were not written out to the config when the PackageDirectory key is missing.
* Renamed "master" branch to "main."


RELEASE.md
```diff
@@ -1,5 +1,15 @@
...
 
+## 1.1.4
+
+- Minumum Runtime Qt SDK version is 100.9.
+- Minimum Qt SDK version is Qt 5.15.0.
+- Minimum Android SDK version is now 21, target SDK version is 29.
+- Fixed erronious "missing file" dialog appearing on startup when opening an MSPK. This was a layer-to-json issue when saving the config.
+- Reverted the "ModelSceneSymbol to PictureMarkerSymbol" change. A camera bug was was introduced because a rotation-expression can not be used in place of a heading-expression when working with an `OrbitGeoElementCameraController`. We will need to park this upgrade until an enhancement is exposed by ArcGISRuntime. 
+- Fixed bug where layers were not written out to the config when the `PackageDirectory` key is missing.
+- Renamed "master" branch to "main."
```
esreli, line 11: After thinking it over, I don't think we should include this line item in the RELEASE notes. Do you feel strongly about keeping this item here?

Shared/qml/ContactReportSizePage.qml
```diff
@@ -18,7 +18,7 @@ import QtQuick 2.6
 import QtQuick.Controls 2.1
 import QtQuick.Controls.Material 2.1
 import QtQuick.Window 2.2
-import Esri.ArcGISRuntime.OpenSourceApps.DSA 1.1
+import Esri.ArcGISRuntime.OpenSourceApps.DSA 1.2
```
esreli, line 21: What is this? It looks oddly similar to the app's version number, should this reflect `1.1.4`?

docs/README.md
```diff
@@ -435,7 +435,7 @@ For convenience, you can try out the DSA apps compiled for [Windows] and [Androi
 
 You can customize aspects of the DSA apps via a configuration file. When the DSA app runs, the app will create a new configuration file if one is not found on local storage. When the app opens the configuration file, the file is read and its values used by the app for that run. Some settings may be changed at run time using the [Settings panel](#settings-panel). Those changes are persisted in the configuration file.
 
-The configuration file is located at `~/ArcGIS/Runtime/Data/DSA/DsaAppConfig.json`, where `~` is `%username%`/`C:/Users/<username>` on Windows, `$HOME` on Unix, and `/sdcard` on Android.
+The configuration file is located at `~/ArcGIS/Runtime/Data/DSA/DsaAppConfig.json`, where `~` is `%username%`/`C:/Users/<username>` on Windows, `$HOME` on Unix and `/sdcard` on Android.
```
esreli, line 438: I checked with the Runtime docs team and they informed me that we follow the Oxford comma rule in all SDK docs. Is there a reason you removed the comma?

Handheld/AppInfo.h
```diff
@@ -22,7 +22,7 @@
 #define kOrganizationDomain             "esri.com"
 
 #define kApplicationName                "DSA_Handheld_Qt"
-#define kApplicationVersion             "1.1.3"
+#define kApplicationVersion             "1.1.4"
 #define kApplicationDescription         "Dynamic Situational Awareness - Handheld app"
```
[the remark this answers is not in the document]
anmacdonald: It's used in the About page of DSA, and only there I think. Many of these `AppInfo` templates are legacy carryover from the early days of AppStudio.

RELEASE.md
anmacdonald: If it were up to me I'd not have the word master anywhere in the repo. I'll remove this line.

Shared/qml/ContactReportSizePage.qml
anmacdonald: Good catch, this should be 1.1 but was missed in the revert. QML version numbers go MAJOR.MINOR with no patch. I'll fix. 

In the long term it doesn't matter too much as it's in the process of being deprecated in Qt 6 though which I am thrilled about. QML file versioning only served to avoiding breaking changes which is resolved with better name resolution in Qt 6. If it were up to me I would force all QML versioning at 1.0; having QML file versions match app versions  is a maintenance nightmare where all files change needlessly with a patch.

docs/README.md
anmacdonald: Just removed the section on the folder locations, which resulted in this change. I'll add it back.

anmacdonald:
@esreli I've addressed your comments. I'll go ahead and take the initiative to merge so I can get the prebuilt binaries out the door today.

PR #350, opened 2020-09-24
Initial Location is written out on each scene switch.

anmacdonald:
InitialLocation is once again written out on startup, and with the addtion of MSPK support whenever the scene changes. The workflow is as follows:

1. User starts up app with no default config file. Immediately the default  config will be written out which contains the Monterey location.
1. The user clicks "`Open` -> Package name -> Scene in Package." The Scene loads.
   * The camera is placed at the loaded scene's initialLocation, the config file's initial location is changed to match.
1. The user closes the app.
1. The user reopens the app and the previous scene is loaded at the initial location in the config file, which matches the scene's.
1. The user closes the app again and manually updates the config file's initial location to Greenland.
1. The user loads the app. The scene is loaded and the camera is placed in Greenland.
1. The user clicks `Open` -> `Reset to Default Scene`.
1. The default scene is loaded. The camera is placed in Monterey. The config file is changed to Monterey.

### DIFF
--- a/Shared/DsaController.cpp
+++ b/Shared/DsaController.cpp
@@ -55,6 +55,8 @@ namespace Dsa {
 
 bool readJsonFile(QIODevice& device, QSettings::SettingsMap& map);
 bool writeJsonFile(QIODevice& device, const QSettings::SettingsMap& map);
+Viewpoint viewpointFromJson(QJsonObject initialLocation);
+QJsonObject defaultViewpoint();
 
 /*!
   \class Dsa::DsaController
@@ -86,14 +88,24 @@ DsaController::DsaController(QObject* parent):
 {
   // setup config settings
   setupConfig();
-  m_scene->setInitialViewpoint(defaultViewpoint());
+  m_scene->setInitialViewpoint(viewpointFromJson(defaultViewpoint()));
   m_dataPath = m_dsaSettings["RootDataDirectory"].toString();
 
   connect(m_scene, &Scene::errorOccurred, this, &DsaController::onError);
 
-  connect(ToolResourceProvider::instance(), &ToolResourceProvider::sceneChanged, this, [this]()
+  connect(ToolResourceProvider::instance(), &ToolResourceProvider::sceneChanged, this, [this, firstLoad{true}]() mutable
   {
     m_scene = ToolResourceProvider::instance()->scene();
+
+    // Here we set up the initial location of the scene fron the config. We do this on the first scene
+    // we load when the application starts up. On any subsequent scenes that the user loads
+    // via the UI, we take this new initial location as the source of truth from then on and
+    // write it to file.
+    if (!m_scene)
+      return;
+
+    updateInitialLocationOnSceneChange(firstLoad);
+    firstLoad = false;
   });
 }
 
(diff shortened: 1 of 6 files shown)


Shared/DsaController.cpp
```diff
@@ -86,14 +88,24 @@ DsaController::DsaController(QObject* parent):
...
+
+    // Here we set up the initial location of the scene fron the config. We do this on the first scene
+    // we load when the application starts up. On any subsequent scenes that the user loads
+    // via the UI, we take this new initial location as the source of truth from then on and
+    // write it to file.
+    if (!m_scene)
+      return;
+
+    updateInitialLocationOnSceneChange(firstLoad);
+    firstLoad = false;
```
lsmallwood, line 108: I don't see where we use `firstLoad` after this call?

Shared/DsaController.cpp
```diff
@@ -568,6 +522,75 @@ void DsaController::saveSettings()
...
+  QJsonObject initialLocationJson;
+  const QString centerString = initialCamera.location().toJson();
+  const QJsonDocument centerDoc = QJsonDocument::fromJson(centerString.toLatin1());
+  initialLocationJson.insert( QStringLiteral("center"), centerDoc.object());
+  initialLocationJson.insert( QStringLiteral("distance"), 0);
+  initialLocationJson.insert( QStringLiteral("heading"), initialCamera.heading());
+  initialLocationJson.insert( QStringLiteral("pitch"), initialCamera.pitch());
+  initialLocationJson.insert( QStringLiteral("roll"), initialCamera.roll());
+
+  m_dsaSettings[QStringLiteral("InitialLocation")] = initialLocationJson.toVariantMap();
```
lsmallwood, line 543: might be nice to have some kind of constant for `"InitialLocation"`

Shared/DsaController.cpp
```diff
@@ -607,6 +630,59 @@ bool writeJsonFile(QIODevice& device, const QSettings::SettingsMap& map)
   return writtenBytes != -1;
 }
 
+Viewpoint viewpointFromJson(QJsonObject initialLocation)
```
lsmallwood, line 633: can this method be `const`? or even `static`?

Shared/packages/OpenMobileScenePackageController.cpp
```diff
@@ -494,6 +495,15 @@ void OpenMobileScenePackageController::updatePackageDetails()
 
 }
 
+/*!
+ * \brief Returns \c true when there is a valid scene package
+ * and scene index cached in this object.
+ */
+bool OpenMobileScenePackageController::hasActiveScene()
```
lsmallwood, line 502: `const`

lsmallwood (approved):
Looks good @anmacdonald - a few comments

Shared/DsaController.cpp
anmacdonald: Only in this call do we use it, it's only of interest whenever the scene changes.

Shared/DsaController.cpp
anmacdonald: This is a freestanding function not in a class method context.

Shared/packages/OpenMobileScenePackageController.cpp
anmacdonald: Good catch!

Shared/DsaController.cpp
lsmallwood: D'oh! I see. Does it need to be in the header or can we put it in an anonymous namespace? If not, how you have it is fine

anmacdonald: Because `writeJsonFile` and `readJsonFile` weren't, I was following convention. But if there's no convention there I'll wrap them all up!

PR #355, opened 2021-03-09
Update guide and reference links in documentation README

mikewilburn:
This PR replaces, updates, or removes links which presently 404 due to the new Developer Site experience.

To simplify the process of reviewing this PR and understand the effect of the change, I've tried to make the commits as atomic as possible. In some cases, you can expand the commit message for more details.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -288,7 +288,7 @@ DSA apps support several generic message feeds.
 
 ## Exploratory visual analysis
 
-Exploratory [analysis tools] combine information such as feature location, your location, and elevation to answer questions about the current operational environment. Analysis is continuously updated as the environment changes, leveraging the GPU of the device and the data that is drawn on the map. The results of an analysis are presented in an overlay that can be [managed like other overlays](#overlays).
+Exploratory analysis tools combine information such as feature location, your location, and elevation to answer questions about the current operational environment. Analysis is continuously updated as the environment changes, leveraging the GPU of the device and the data that is drawn on the map. The results of an analysis are presented in an overlay that can be [managed like other overlays](#overlays).
 
 ### Viewshed
 
@@ -308,7 +308,7 @@ Exploratory [analysis tools] combine information such as feature location, your 
 
 ***Developer tips:***
 
-- Both viewshed and line of sight analysis are calculated using the GPU and operate only on the data displayed on the map. This means that the accuracy of these analyses are limited by the current resolution of the displayed data and the elevation surface. More information can be found in the section [Scene analyses versus geoprocessing] in the Guide topic Analyze visibility in a scene view.
+- Both viewshed and line of sight analysis are calculated using the GPU and operate only on the data displayed on the map. This means that the accuracy of these analyses are limited by the current resolution of the displayed data and the elevation surface.
 
 ## Alerts and conditions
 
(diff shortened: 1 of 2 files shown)


mikewilburn:
Hey @kerryrobinson & @esreli, I'm still looking for a few more replacement links, but feel free to begin reviewing and share any feedback on the changes I've made thus far.

Before I merge, I also plan to update the `RELEASE.md` doc and increment the release version number with a patch to account for this change. I'll eventually add a commit which accounts for that change inside of this PR.

docs/README.md
```diff
@@ -215,22 +215,22 @@ Some [overlays](#overlays) get their information from local data sources, such a
 
 Specify the folders the app searches for local data using your [App configuration settings](#app-configuration-settings). Local data that has been found on your device is listed by this tool. Any data supported by ArcGIS Runtime SDKs can be used, including:
 
-- [GeoPackage]
+- [Geopackage]
 - [Tile package] (TPK)
 - [Mobile geodatabase]
 - KML
```
kerryrobinson, line 221: I feel like we might as well add a link to KML in the glossary as well: https://developers.arcgis.com/documentation/glossary/kml/

kerryrobinson (commented):
@mikewilburn so far the changes look good to me. frustrating some of the doc that is no longer in the guide that added some helpful context to developers.

docs/README.md
mikewilburn: Good suggestion - resolved with d0dae19

mikewilburn:
@kerryrobinson completely agree. I think unfortunately, broken/removed links affecting this doc might be the norm and not the exception 😬 . Addressed that one comment you made - I think all should be good now.

@esreli I believe I've found and resolved all of the broken links now. This is ready for your review.

docs/README.md
```diff
@@ -122,7 +122,7 @@ The navigation toolbar can be turned off in the [Settings panel](#settings-panel
 
 There is a compass on the display's lower-right corner. Its red pointer points to north. Touching the compass will rotate the map so the top points toward north.
 
-The [compass tool] is part of ArcGIS Runtime Toolkit for Qt.
+The [NorthArrow compass control] is part of ArcGIS Runtime Toolkit for Qt.
```
esreli, line 125: Perhaps we link users to the Toolkit from this line?

esreli (commented):
Great job capturing all of the dead links. I only have one general question and one in-line suggestion.

- There is a commented out image in the Compass section (line `120`). Can we resolve this?

mikewilburn:
> - There is a commented out image in the Compass section (line 120). Can we resolve this?

Good spot. It's unclear to me why that was there. A little digging revealed that line 110 was the appropriate place for that, which is already the case. I've removed line 120 altogether.

docs/README.md
mikewilburn: I've left the specific reference to the NorthArrow component while still supplementing this line with a link to the Toolkit repo. Hopefully this is what you meant. I do think both links serve a purpose here.

mikewilburn:
@esreli this PR now also includes an update to the `RELEASE.md` doc to describe the new documentation changes.

esreli (approved):
LGTM 👍

PR #360, opened 2021-06-11
Update formatting of README.md to accommodate Page layout

kerryrobinson:
I smushed the TOC into a few rows, which I think reads a little better and should work well with the Page format.


ldanzinger (approved):
LGTM

PR #363, opened 2022-12-19
Luca6804/200

ldanzinger:
@JamesMBallard please review

### DIFF
--- a/Handheld/Android/AndroidManifest.xml
+++ b/Handheld/Android/AndroidManifest.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <manifest package="com.esri.arcgisruntime.opensourceapps.DSA_Handheld_Qt" xmlns:android="http://schemas.android.com/apk/res/android" android:versionName="1.1.6" android:versionCode="7" android:installLocation="auto">
-    <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="29"/>
+    <uses-sdk android:minSdkVersion="26" android:targetSdkVersion="29"/>
     <!-- The following comment will be replaced upon deployment with default permissions based on the dependencies of the application.
          Remove the comment if you do not require these default permissions. -->
     <!-- %%INSERT_PERMISSIONS -->
--- a/Shared/alerts/AlertSource.h
+++ b/Shared/alerts/AlertSource.h
@@ -17,8 +17,9 @@
 #ifndef ALERTSOURCE_H
 #define ALERTSOURCE_H
 
-// C++ API headers
-#include "Point.h"
+namespace Esri::ArcGISRuntime {
+class Point;
+}
 
 // Qt headers
 #include <QObject>
--- a/Shared/alerts/AlertSource.h
+++ b/Shared/alerts/AlertSource.h
@@ -18,7 +18,7 @@
 #define ALERTSOURCE_H
 
 namespace Esri::ArcGISRuntime {
-class Point;
+  class Point;
 }
 
 // Qt headers
--- a/Shared/messages/ObservationReportController.cpp
+++ b/Shared/messages/ObservationReportController.cpp
@@ -367,13 +367,13 @@ void ObservationReportController::onMouseClicked(QMouseEvent& event)
   SceneView* sceneView = dynamic_cast<SceneView*>(m_geoView);
   if (sceneView)
   {
-    setControlPoint(sceneView->screenToBaseSurface(event.x(), event.y()));
+    setControlPoint(sceneView->screenToBaseSurface(event.pos().x(), event.pos().y()));
   }
   else
   {
     MapView* mapView = dynamic_cast<MapView*>(m_geoView);
     if (mapView)
-      setControlPoint(mapView->screenToLocation(event.x(), event.y()));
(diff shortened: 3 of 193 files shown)


Shared/alerts/AlertSource.h
```diff
@@ -17,8 +17,9 @@
 #ifndef ALERTSOURCE_H
 #define ALERTSOURCE_H
 
-// C++ API headers
-#include "Point.h"
+namespace Esri::ArcGISRuntime {
+class Point;
```
JamesMBallard, line 21: Minor nitpick. Some of these are indented and others aren't. Choose one (my vote is indented). If this will add more time and codebase churn, leave it.

Shared/messages/ObservationReportController.cpp
```diff
@@ -367,13 +367,13 @@ void ObservationReportController::onMouseClicked(QMouseEvent& event)
   SceneView* sceneView = dynamic_cast<SceneView*>(m_geoView);
   if (sceneView)
   {
-    setControlPoint(sceneView->screenToBaseSurface(event.x(), event.y()));
+    setControlPoint(sceneView->screenToBaseSurface(event.pos().x(), event.pos().y()));
```
JamesMBallard, line 370: ditto using `position()`. I think that's correct, but the types are different. We need to validate that's correct still.

Shared/packages/MobileScenePackagesListModel.cpp
```diff
@@ -195,7 +195,7 @@ QVariant MobileScenePackagesListModel::data(const QModelIndex& index, int role)
     return QVariant();
 
   auto it = m_packageDetails.constBegin();
-  it.operator +=(index.row());
+  std::advance(it, index.row());
```
JamesMBallard, line 198: I recommend adding `#include <iterator>` in this file to be safe.

https://en.cppreference.com/w/cpp/iterator/advance

Vehicle/Android/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <manifest package="com.esri.arcgisruntime.opensourceapps.DSA_Vehicle_Qt" xmlns:android="http://schemas.android.com/apk/res/android" android:versionName="1.1.6" android:versionCode="7" android:installLocation="auto">
-    <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="29"/>
+    <uses-sdk android:minSdkVersion="26" android:targetSdkVersion="29"/>
     <!-- The following comment will be replaced upon deployment with default permissions based on the dependencies of the application.
          Remove the comment if you do not require these default permissions. -->
     <!-- %%INSERT_PERMISSIONS -->
```
JamesMBallard, line 6: Ditto entries below need updating to Qt 6.

Vehicle/AppInfo.h
```diff
@@ -22,7 +22,7 @@
 #define kOrganizationDomain             "esri.com"
 
 #define kApplicationName                "DSA_Vehicle_Qt"
-#define kApplicationVersion             "1.1.6"
+#define kApplicationVersion             "2.0.0"
```
JamesMBallard, line 25: Is there some DSA versioning we should follow? 2.0.0 seems like a good choice.

Handheld/Android/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <manifest package="com.esri.arcgisruntime.opensourceapps.DSA_Handheld_Qt" xmlns:android="http://schemas.android.com/apk/res/android" android:versionName="1.1.6" android:versionCode="7" android:installLocation="auto">
-    <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="29"/>
+    <uses-sdk android:minSdkVersion="26" android:targetSdkVersion="29"/>
     <!-- The following comment will be replaced upon deployment with default permissions based on the dependencies of the application.
          Remove the comment if you do not require these default permissions. -->
     <!-- %%INSERT_PERMISSIONS -->
```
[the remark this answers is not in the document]
ldanzinger: I was planning on working on the Android porting as a separate PR since the permissions and data location will need to be updated like we did in samples...

Vehicle/AppInfo.h
ldanzinger: Not that I'm aware of. I suppose you could make an argument for this just getting updated to 1.2.0 since the app itself doesn't have much new functionality. Maybe that would make more sense.

Vehicle/Android/AndroidManifest.xml
ldanzinger: I'll update the manifest and the permissions stuff in another PR

Shared/packages/MobileScenePackagesListModel.cpp
ldanzinger: yup, it's there already 😄

Vehicle/AppInfo.h
JamesMBallard: I think 2.0.0 makes sense. It brings it up to both Qt 6 and Qt Maps SDK 200.0, which are both major-version releases.

JamesMBallard (approved):
Either version works. We can wait for a 2.0 once we update some of the features to user newer Maps SDK features.

Issue #367, opened 2023-09-01
Update real time feeds to make use of dynamic entities custom data source


ldanzinger:
# Summary

DSA simulates real time data feeds with a message simulator that sends GeoMessage and Cursor On Target messages. The DSA app listens to a UDP port and displays the messages as graphics on the screen. If the message is an update to an existing graphic, the graphic attributes and geometry will be updated.

DSA was written before Runtime/Maps SDK had official support for real time feeds. Since then, we've added dynamic entities, which allow you to hook into stream services, as well as custom data sources.

We should update DSA to make use of custom data sources to use official best practices instead of the hand-rolled graphics implementation. In the end, the message feeds should be represented by DynamicEntityLayer with DynamicEntityDataSource instead of GraphicsOverlay.

All work should be done off of the v.next branch, which will build with Qt 6.5.1 and Maps SDK 200.2.

# Definition of done

- [x] App is updated to use DynamicEntityLayer for displaying real time feeds
- [x] Updates work for GeoMessage and Cursor On Target
- [ ] All other functions in the app continue to work as they did previously, such as Alerts

kerryrobinson:
Instead of creating a new issue, I'll add a comment here on issues/bugs I've noticed with the update to use Dynamic Entities, since it is not merged yet.  We can determine if we need to turn any of these into new issues. 

1- The `Message Feeds` panel lists duplicates of the 7 feeds. The bottom set of feeds in the list do not have any effect on turning on or off the visibility of the feeds (only the top set).  I noticed that this only happens the first time I open the app and if there hasn't already been a config file created. Once the config file is created, I cannot reproduce this issue. 
![image](https://github.com/Esri/dynamic-situational-awareness-qt/assets/3383476/71c3b497-3293-431b-8a86-806e76d607c0)

2 - Dynamic Entities layers show up in the `Overlays` panel, when they should only appear in the `Message Feeds` panel. (note these are also duplicated when the app opens for the first time without an existing config file present)
![image](https://github.com/Esri/dynamic-situational-awareness-qt/assets/3383476/d47391fd-0be5-4c42-bac4-ff99f0c1fdcf)

3 - Long-pressing on a dynamic entity in a feed results in an app crash

4- The Conditions list says there are no conditions, even though there is one condition saved in the config file (the default condition to highlight any Land Friendly Tracks that have a status of 1). 

[...]

PR #371, opened 2024-01-10
Dtedder/367 update feeds to dynamic entities

dtedder:
@JamesMBallard please review.

Main points are:
- converted MessageFeed into a DynamicEntityDataSource
- switched MessagesOverlay to use DynamicEntityLayer instead of GraphicsOverlay
- added a new AlertTarget type for DynamicEntityLayer
- adjusted Alerts controller to look for DynamicEntityLayer types in the map
- added a new AlertSource type for DynamicEntity

### DIFF
--- a/Handheld/Handheld.pro
+++ b/Handheld/Handheld.pro
@@ -20,7 +20,7 @@ TEMPLATE = app
 QT += core gui opengl network positioning sensors qml quick xml
 CONFIG += c++17
 
-ARCGIS_RUNTIME_VERSION = 200.2.0
+ARCGIS_RUNTIME_VERSION = 200.4.0
 include($$PWD/../Shared/build/arcgisruntime.pri)
 
 INCLUDEPATH += $$PWD/../Shared/ \
--- a/Shared/DrawOrderLayerListModel.h
+++ b/Shared/DrawOrderLayerListModel.h
@@ -32,6 +32,7 @@ public:
 
 protected:
   bool lessThan(const QModelIndex& sourceLeft, const QModelIndex& sourceRight) const override;
+  bool filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const override;
 };
 
 } // Dsa
--- a/Shared/DrawOrderLayerListModel.h
+++ b/Shared/DrawOrderLayerListModel.h
@@ -32,7 +32,7 @@ public:
 
 protected:
   bool lessThan(const QModelIndex& sourceLeft, const QModelIndex& sourceRight) const override;
-  bool filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const override;
+  bool filterAcceptsRow(int sourceRow, const QModelIndex& sourceParent) const override;
 };
 
 } // Dsa
--- a/Shared/alerts/AlertCondition.cpp
+++ b/Shared/alerts/AlertCondition.cpp
@@ -145,17 +145,9 @@ void AlertCondition::init(Esri::ArcGISRuntime::DynamicEntityLayer* sourceFeed, c
   connect(dataSource, &DynamicEntityDataSource::dynamicEntityReceived, this, [this, sourceFeed, target](DynamicEntityInfo* info)
   {
     auto* dynamicEntity = info->dynamicEntity();
-    dynamicEntity->setParent(this);
     DynamicEntityAlertSource* source = new DynamicEntityAlertSource(dynamicEntity, sourceFeed);
     AlertConditionData* newData = createData(source, target);
     addData(newData);
-    info->deleteLater();
-  });
-
-  connect(dataSource, &DynamicEntityDataSource::dynamicEntityPurged, this, [](DynamicEntityInfo* info)
-  {
-    auto* dynamicEntity = info->dynamicEntity();
-    dynamicEntity->setParent(nullptr);
     info->deleteLater();
   });
 }
--- a/Shared/alerts/AlertCondition.cpp
+++ b/Shared/alerts/AlertCondition.cpp
@@ -18,14 +18,21 @@
 #include "pch.hpp"
 
 #include "AlertCondition.h"
+#include "AlertListModel.h"
 
 // dsa app headers
 #include "AlertConditionData.h"
 #include "GraphicAlertSource.h"
+#include "DynamicEntityAlertSource.h"
 
 // C++ API headers
+#include "Graphic.h"
 #include "GraphicListModel.h"
 #include "GraphicsOverlay.h"
+#include "DynamicEntity.h"
+#include "DynamicEntityInfo.h"
+#include "DynamicEntityLayer.h"
+#include "DynamicEntityDataSource.h"
 
 using namespace Esri::ArcGISRuntime;
 
@@ -124,6 +131,33 @@ void AlertCondition::init(GraphicsOverlay* sourceFeed, const QString& sourceDesc
   const int count = graphics->rowCount();
   for (int i = 0; i < count; ++i)
     handleGraphicAt(i);
+}
+
+void AlertCondition::init(Esri::ArcGISRuntime::DynamicEntityLayer* sourceFeed, const QString& sourceDescription, AlertTarget* target, const QString& targetDescription)
+{
+  if (!sourceFeed || !target)
+    return;
+
+  m_sourceDescription = sourceDescription;
+  m_targetDescription = targetDescription;
+  const auto* dataSource = sourceFeed->dataSource();
+
+  connect(dataSource, &DynamicEntityDataSource::dynamicEntityReceived, this, [this, sourceFeed, target](DynamicEntityInfo* info)
+  {
+    auto* dynamicEntity = info->dynamicEntity();
+    dynamicEntity->setParent(this);
+    DynamicEntityAlertSource* source = new DynamicEntityAlertSource(dynamicEntity, sourceFeed);
+    AlertConditionData* newData = createData(source, target);
+    addData(newData);
+    info->deleteLater();
+  });
+
+  connect(dataSource, &DynamicEntityDataSource::dynamicEntityPurged, this, [](DynamicEntityInfo* info)
+  {
+    auto* dynamicEntity = info->dynamicEntity();
+    dynamicEntity->setParent(nullptr);
+    info->deleteLater();
+  });
 }
 
 /*!
--- a/Shared/alerts/AlertCondition.cpp
+++ b/Shared/alerts/AlertCondition.cpp
@@ -29,7 +29,6 @@
 #include "Graphic.h"
 #include "GraphicListModel.h"
 #include "GraphicsOverlay.h"
-#include "DynamicEntity.h"
 #include "DynamicEntityInfo.h"
 #include "DynamicEntityLayer.h"
 #include "DynamicEntityDataSource.h"
@@ -133,7 +132,7 @@ void AlertCondition::init(GraphicsOverlay* sourceFeed, const QString& sourceDesc
     handleGraphicAt(i);
 }
 
-void AlertCondition::init(Esri::ArcGISRuntime::DynamicEntityLayer* sourceFeed, const QString& sourceDescription, AlertTarget* target, const QString& targetDescription)
+void AlertCondition::init(DynamicEntityLayer* sourceFeed, const QString& sourceDescription, AlertTarget* target, const QString& targetDescription)
 {
   if (!sourceFeed || !target)
     return;
--- a/Shared/alerts/AlertConditionsController.cpp
+++ b/Shared/alerts/AlertConditionsController.cpp
@@ -406,24 +406,29 @@ bool AlertConditionsController::addAttributeEqualsAlert(const QString& condition
     return false;
   }
 
-  AlertTarget* target = new FixedValueAlertTarget(targetValue, this);
+  GraphicsOverlay* sourceGraphicsOverlay = graphicsOverlayFromName(sourceFeedName);
+  DynamicEntityLayer* sourceDynamicEntityLayer = nullptr;
+  if (!sourceGraphicsOverlay)
+  {
+    sourceDynamicEntityLayer = dynamicEntityLayerFromName(sourceFeedName);
+  }
+  if (!sourceGraphicsOverlay && !sourceDynamicEntityLayer)
+  {
+    emit toolErrorOccurred(QStringLiteral("Failed to create Condition"), QString("Could not find source feed: %1").arg(sourceFeedName));
+    return false;
+  }
 
+  AlertTarget* target = new FixedValueAlertTarget(targetValue, this);
   AttributeEqualsAlertCondition* condition = new AttributeEqualsAlertCondition(level, conditionName, attributeName, this);
   connect(condition, &AttributeEqualsAlertCondition::newConditionData, this, &AlertConditionsController::handleNewAlertConditionData);
 
-  if (GraphicsOverlay* sourceOverlay = graphicsOverlayFromName(sourceFeedName); sourceOverlay)
+  if (sourceGraphicsOverlay)
   {
-    condition->init(sourceOverlay, sourceFeedName, target, targetValue.toString());
-  }
-  else if (DynamicEntityLayer* dynamicEntityLayer = dynamicEntityLayerFromName(sourceFeedName); dynamicEntityLayer)
-  {
-    condition->init(dynamicEntityLayer, sourceFeedName, target, targetValue.toString());
+    condition->init(sourceGraphicsOverlay, sourceFeedName, target, targetValue.toString());
   }
   else
   {
-    delete target;
-    emit toolErrorOccurred(QStringLiteral("Failed to create Condition"), QString("Could not find source feed: %1").arg(sourceFeedName));
-    return false;
+    condition->init(sourceDynamicEntityLayer, sourceFeedName, target, targetValue.toString());
   }
   return m_conditions->addAlertCondition(condition);
 }
--- a/Shared/alerts/AlertConditionsController.cpp
+++ b/Shared/alerts/AlertConditionsController.cpp
@@ -609,25 +609,24 @@ void AlertConditionsController::onLayersChanged()
       if (!lyr)
         continue;
 
-      FeatureLayer* featLayer = qobject_cast<FeatureLayer*>(lyr);
-      DynamicEntityLayer *dynamicEntityLayer = qobject_cast<DynamicEntityLayer*>(lyr);
-      if (!featLayer && !dynamicEntityLayer)
-        continue;
-
-      if (featLayer)
+      if (FeatureLayer* featureLayer = qobject_cast<FeatureLayer*>(lyr); featureLayer)
       {
-        if (featLayer->loadStatus() != LoadStatus::Loaded)
-          connect(featLayer, &FeatureLayer::doneLoading, this, &AlertConditionsController::onLayersChanged);
+        if (featureLayer->loadStatus() != LoadStatus::Loaded)
+        {
+          connect(featureLayer, &FeatureLayer::doneLoading, this, &AlertConditionsController::onLayersChanged);
+        }
         else
         {
-          newTargetList.append(featLayer->name());
-          existingLayerIds.append(featLayer->name());
+          newTargetList.append(featureLayer->name());
+          existingLayerIds.append(featureLayer->name());
         }
       }
-      if (dynamicEntityLayer)
+      else if (DynamicEntityLayer* dynamicEntityLayer = qobject_cast<DynamicEntityLayer*>(lyr); dynamicEntityLayer)
       {
         if (dynamicEntityLayer->loadStatus() != LoadStatus::Loaded)
+        {
           connect(dynamicEntityLayer, &DynamicEntityLayer::doneLoading, this, &AlertConditionsController::onLayersChanged);
+        }
         else
         {
           newTargetList.append(dynamicEntityLayer->name());
@@ -1089,46 +1088,41 @@ AlertTarget* AlertConditionsController::targetFromItemIdAndIndex(int itemId, int
       if (!layer)
         continue;
 
-      FeatureLayer *featLayer = qobject_cast<FeatureLayer*>(layer);
-      DynamicEntityLayer *dynamicEntityLayer = qobject_cast<DynamicEntityLayer*>(layer);
-      if (!featLayer && !dynamicEntityLayer)
-        continue;
-
       currIndex++;
 
       if (currIndex == targetOverlayIndex)
       {
-          if (featLayer)
+        if (FeatureLayer* featureLayer = qobject_cast<FeatureLayer*>(layer); featureLayer)
+        {
+          if (itemId == -1)
           {
-            if (itemId == -1)
-            {
-              if (!m_layerTargets.contains(featLayer->name()))
-                m_layerTargets.insert(featLayer->name(), new FeatureLayerAlertTarget(featLayer));
+            if (!m_layerTargets.contains(featureLayer->name()))
+              m_layerTargets.insert(featureLayer->name(), new FeatureLayerAlertTarget(featureLayer));
 
-              targetDescription = featLayer->name();
-              return m_layerTargets.value(featLayer->name(), nullptr);
-            }
-            else
-            {
-              targetDescription = QString("%1 [%2]").arg(featLayer->name(), QString::number(itemId));
-              return targetFromFeatureLayer(featLayer, itemId);
-            }
+            targetDescription = featureLayer->name();
+            return m_layerTargets.value(featureLayer->name(), nullptr);
           }
-          if (dynamicEntityLayer)
+          else
           {
-              if (itemId == -1)
-              {
-                  if (!m_layerTargets.contains(dynamicEntityLayer->name()))
-                      m_layerTargets.insert(dynamicEntityLayer->name(), new DynamicEntityLayerAlertTarget(dynamicEntityLayer));
-
-                  targetDescription = dynamicEntityLayer->name();
-                  return m_layerTargets.value(dynamicEntityLayer->name(), nullptr);
-              }
-              else
-              {
-                  return nullptr;
-              }
+            targetDescription = QString("%1 [%2]").arg(featureLayer->name(), QString::number(itemId));
+            return targetFromFeatureLayer(featureLayer, itemId);
           }
+        }
+        else if (DynamicEntityLayer* dynamicEntityLayer = qobject_cast<DynamicEntityLayer*>(layer); dynamicEntityLayer)
+        {
+          if (itemId == -1)
+          {
+            if (!m_layerTargets.contains(dynamicEntityLayer->name()))
+              m_layerTargets.insert(dynamicEntityLayer->name(), new DynamicEntityLayerAlertTarget(dynamicEntityLayer));
+
+            targetDescription = dynamicEntityLayer->name();
+            return m_layerTargets.value(dynamicEntityLayer->name(), nullptr);
+          }
+          else
+          {
+            return nullptr;
+          }
+        }
       }
     }
   }
--- a/Shared/alerts/AlertConditionsController.cpp
+++ b/Shared/alerts/AlertConditionsController.cpp
@@ -108,8 +108,8 @@ AlertConditionsController::AlertConditionsController(QObject* parent /* = nullpt
 {
   connect(ToolResourceProvider::instance(), &ToolResourceProvider::geoViewChanged,
           this, &AlertConditionsController::onGeoviewChanged);
-    connect(ToolResourceProvider::instance(), &ToolResourceProvider::sceneChanged,
-          this, &AlertConditionsController::onGeoviewChanged);
+  connect(ToolResourceProvider::instance(), &ToolResourceProvider::sceneChanged,
+        this, &AlertConditionsController::onGeoviewChanged);
 
   connect(m_conditions, &AlertConditionListModel::rowsInserted, this, &AlertConditionsController::onConditionsChanged);
   connect(m_conditions, &AlertConditionListModel::rowsRemoved, this, &AlertConditionsController::onConditionsChanged);
@@ -616,23 +616,23 @@ void AlertConditionsController::onLayersChanged()
 
       if (featLayer)
       {
-          if (featLayer->loadStatus() != LoadStatus::Loaded)
-              connect(featLayer, &FeatureLayer::doneLoading, this, &AlertConditionsController::onLayersChanged);
-          else
-          {
-              newTargetList.append(featLayer->name());
-              existingLayerIds.append(featLayer->name());
-          }
+        if (featLayer->loadStatus() != LoadStatus::Loaded)
+          connect(featLayer, &FeatureLayer::doneLoading, this, &AlertConditionsController::onLayersChanged);
+        else
+        {
+          newTargetList.append(featLayer->name());
+          existingLayerIds.append(featLayer->name());
+        }
       }
       if (dynamicEntityLayer)
       {
-          if (dynamicEntityLayer->loadStatus() != LoadStatus::Loaded)
-              connect(dynamicEntityLayer, &DynamicEntityLayer::doneLoading, this, &AlertConditionsController::onLayersChanged);
-          else
-          {
-              newTargetList.append(dynamicEntityLayer->name());
-              existingLayerIds.append(dynamicEntityLayer->name());
-          }
+        if (dynamicEntityLayer->loadStatus() != LoadStatus::Loaded)
+          connect(dynamicEntityLayer, &DynamicEntityLayer::doneLoading, this, &AlertConditionsController::onLayersChanged);
+        else
+        {
+          newTargetList.append(dynamicEntityLayer->name());
+          existingLayerIds.append(dynamicEntityLayer->name());
+        }
       }
     }
   }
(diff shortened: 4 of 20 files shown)


Shared/alerts/AlertConditionsController.cpp
```diff
@@ -606,15 +610,29 @@ void AlertConditionsController::onLayersChanged()
...
-        existingLayerIds.append(featLayer->name());
+        if (featLayer->loadStatus() != LoadStatus::Loaded)
+          connect(featLayer, &FeatureLayer::doneLoading, this, &AlertConditionsController::onLayersChanged);
+        else
+        {
+          newTargetList.append(featLayer->name());
+          existingLayerIds.append(featLayer->name());
+        }
+      }
+      if (dynamicEntityLayer)
```
JamesMBallard, line 627: I recommend some minor cleanup here. Rather than qobject_cast'int to both types upfront, check them sequentially in if-else cases. It can't be both. I would also use if-assignment checks to reduce extra scope.

Here's an example
```c++
if (FeatureLayer* featLayer = qobject_cast<FeatureLayer*>(lyr); featLayer)
{
  // do feature layer stuff
}
else if (DynamicEntityLayer* dynamicEntityLayer = qobject_cast<DynamicEntityLayer*>(lyr); dynamicEntityLayer)
{
  // do DynamicEntityLayer stuff
}
else 
{
  continue; // is this still needed here? Please review
}
```

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -1071,27 +1089,46 @@ AlertTarget* AlertConditionsController::targetFromItemIdAndIndex(int itemId, int
...
+            }
+            else
+            {
+              targetDescription = QString("%1 [%2]").arg(featLayer->name(), QString::number(itemId));
+              return targetFromFeatureLayer(featLayer, itemId);
+            }
+          }
+          if (dynamicEntityLayer)
+          {
+              if (itemId == -1)
```
JamesMBallard, line 1119: Please use consistent indent size within the project.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+{
+  return QVariant{};
+}
+
+/*!
+  \internal
+
+  Connect signals etc. for a new \a dynamicEntity.
+ */
+void DynamicEntityLayerAlertTarget::connectEntityGraphic(Esri::ArcGISRuntime::DynamicEntity *dynamicEntity)
```
JamesMBallard, line 142: Nitpick, but in the .cpp files we typically are `using namespace Esri::ArcGISRuntime`, so you don't need to fully qualify here.
```suggestion
void DynamicEntityLayerAlertTarget::connectEntityGraphic(DynamicEntity* dynamicEntity)
```

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+{
+  // check for an existing graphic
+  Graphic *graphic;
```
JamesMBallard, line 145: Another nitpick, but prefer `Type* var` rather than `Type *var`. We want the look and feel to be consistent throughout the codebase.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+  if (m_entityGraphics.contains(dynamicEntity->entityId()))
+  {
+    // update the geometry
+    graphic = m_entityGraphics[dynamicEntity->entityId()];
+    graphic->setGeometry(dynamicEntity->geometry());
+  }
+  else
+  {
+      // if no graphic existed for the entity id, construct a new one and insert it into the graphics lookup
+    graphic = new Graphic(dynamicEntity->geometry());
```
JamesMBallard, line 155: Who owns this graphic? It either needs a parent, or some other type of memory management. Please review any "new" calls without a parent provided.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
+/*******************************************************************************
+ *  Copyright 2012-2018 Esri
```
JamesMBallard, line 2:
```suggestion
 *  Copyright 2012-2024 Esri
```

New files should get new dates. Check throughout please.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+  // destroy the quadtree if it already existed
+  if (m_quadtree)
+  {
+    delete m_quadtree;
+    m_quadtree = nullptr;
+  }
+
+  // build a list of pointers to geoelements for every graphic in the lookup
+  QList<GeoElement*> elements;
+  for (auto const dynamicEntity : std::as_const(m_entityGraphics))
```
JamesMBallard, line 187: Where possible, use `auto*` for pointer types. The Qt SDK does have value types, so I think it helps to convey better what's happening.

Also, prefer `const auto` rather than `auto const`.  We're ["West Const"](https://hackingcpp.com/cpp/design/east_vs_west_const.html) for better or worse.

Shared/alerts/DynamicEntityLayerAlertTarget.h
```diff
@@ -0,0 +1,57 @@
...
+  QList<Esri::ArcGISRuntime::Geometry> targetGeometries(const Esri::ArcGISRuntime::Envelope &targetArea) const override;
+  QVariant targetValue() const override;
+
+private:
+  Esri::ArcGISRuntime::DynamicEntityLayer *m_dynamicEntityLayer = nullptr;
+  GeometryQuadtree *m_quadtree = nullptr;
+  QMap<quint64, Esri::ArcGISRuntime::Graphic*> m_entityGraphics;
+  void connectEntityGraphic(Esri::ArcGISRuntime::DynamicEntity *dynamicEntity);
+  void rebuildQuadtree();
+};
```
JamesMBallard, line 53:
```suggestion
public:
  explicit DynamicEntityLayerAlertTarget(Esri::ArcGISRuntime::DynamicEntityLayer* dynamicEntityLayer);
  ~DynamicEntityLayerAlertTarget();

  QList<Esri::ArcGISRuntime::Geometry> targetGeometries(const Esri::ArcGISRuntime::Envelope& targetArea) const override;
  QVariant targetValue() const override;

private:
  Esri::ArcGISRuntime::DynamicEntityLayer* m_dynamicEntityLayer = nullptr;
  GeometryQuadtree* m_quadtree = nullptr;
  QMap<quint64, Esri::ArcGISRuntime::Graphic*> m_entityGraphics;
  void connectEntityGraphic(Esri::ArcGISRuntime::DynamicEntity* dynamicEntity);
  void rebuildQuadtree();
};
```

Shared/messages/MessageFeed.cpp
```diff
@@ -17,44 +17,143 @@
...
+  // listen for new entities
+  connect(this, &DynamicEntityDataSource::dynamicEntityReceived, this, [this](DynamicEntityInfo *info)
+  {
+    // checck new entity for select/unselect action
+    auto dynamicEntity = info->dynamicEntity();
+    checkEntityForSelectAction(dynamicEntity);
+
+    // mark the info as delete later so it can be cleaned up
+    // info in the source cannot be cleaned up immediately since alert targets might need a reference to the related DynamicEntity
+    info->deleteLater();
```
JamesMBallard, line 128: 👍 this should keep the memory footprint from growing throughout the run.

Shared/messages/MessageFeed.cpp
```diff
@@ -145,9 +230,86 @@ QUrl MessageFeed::thumbnailUrl() const
...
+  }
+
+  const auto symbolId = message.symbolId();
+  const auto geometry = message.geometry();
+  const auto messageAction = message.messageAction();
+
+  switch (messageAction)
+  {
+  case Message::MessageAction::Remove:
+    this->deleteEntityAsync(messageId).then([]{});
```
JamesMBallard, line 264: Is any .then required here? If there's nothing to do, is it enough to just call it? If you get a warning because of unused return type, you can do this:

```suggestion
    auto future = this->deleteEntityAsync(messageId);
    Q_UNUSED(future)
```

That would better convey the intent here IMO.

Shared/messages/MessageFeed.cpp
```diff
@@ -145,9 +230,86 @@ QUrl MessageFeed::thumbnailUrl() const
...
+{
+  // check for selection on add for geomessage types only
+  if (!m_isCoT)
+  {
+    // find the action attribute
+    if (dynamicEntity)
+    {
+      auto actionValue = dynamicEntity->attributes()->attributesMap()[Message::GEOMESSAGE_ACTION_NAME].toString();
+      static QString selectValue = Message::fromMessageAction(Message::MessageAction::Select);
+      static QString unselectValue = Message::fromMessageAction(Message::MessageAction::Unselect);
```
JamesMBallard, line 302:
```suggestion
      static const QString selectValue = Message::fromMessageAction(Message::MessageAction::Select);
      static const QString unselectValue = Message::fromMessageAction(Message::MessageAction::Unselect);
```

Shared/messages/MessagesOverlay.cpp
```diff
@@ -53,33 +48,31 @@ namespace Dsa {
...
   \brief Constructor taking a \a geoView, a \a renderer, a \a messageType,
   a \a surfacePlacement mode and an optional \a parent.
  */
-MessagesOverlay::MessagesOverlay(GeoView* geoView, Renderer* renderer, const QString& messageType,
-                                 SurfacePlacement surfacePlacement, QObject* parent) :
+
+/*!
+  \brief Destructor.
+ */
+MessagesOverlay::MessagesOverlay(Esri::ArcGISRuntime::GeoView *geoView, Esri::ArcGISRuntime::Renderer *renderer, const QString &messageType, MessageFeed *messageFeed, Esri::ArcGISRuntime::SurfacePlacement surfacePlacement, QObject *parent) :
```
JamesMBallard, line 60: You likely don't need the fully qualifies namespaces here.

Also, Qt Creator has opinions about where the "*" and "&" go that don't jive with the rest of the project.

Shared/messages/MessagesOverlay.h
```diff
@@ -70,12 +70,12 @@ class MessagesOverlay : public QObject
...
 
-  Esri::ArcGISRuntime::GeoView* m_geoView = nullptr;
+  Esri::ArcGISRuntime::GeoView *m_geoView = nullptr;
   QPointer<Esri::ArcGISRuntime::Renderer> m_renderer;
   Esri::ArcGISRuntime::SurfacePlacement m_surfacePlacement;
 
-  Esri::ArcGISRuntime::GraphicsOverlay* m_graphicsOverlay = nullptr;
-  QHash<QString, Esri::ArcGISRuntime::Graphic*> m_existingGraphics;
+  Esri::ArcGISRuntime::DynamicEntityLayer *m_dynamicEntityLayer = nullptr;
+  MessageFeed *m_messageFeed;
```
JamesMBallard, line 78:
```suggestion
  MessageFeed* m_messageFeed = nullptr;
```

JamesMBallard (commented):
Looks pretty good @dtedder. A few things to review.

Shared/alerts/AlertConditionsController.cpp
dtedder: I replaced this block as well as another very similar instance later in the code.

Shared/alerts/AlertConditionsController.cpp
dtedder: Disabled clang-formatting and using Qt team style file.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: Forgot to clean these up from the Refactor > AddDefinition to the class where the using namespace makes the full qualifier unnecessary.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: Noted. I'll use this style from now on.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: Wasn't sure what the best practice is for objects like this that are not added to something in the UI like a layer. Graphics created here are deleted via unique_ptr when they are removed from the QMap on the DynamicEntityPurged signal. Should they be given 'this' as their owner in the constructor?

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: The new DynamicEntityLayerAlertTarget.h and .cpp files have been updated to 2024. Should the other's I've modified in this PR be updated as well?

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: I hadn't looked at this style point yet but I think I agree now that having the * is better for clarity.

My use of std::as_const here originated from a clazy warning range-loop-detach. Not sure which is preferred here but I switched to qAsConst as indicated in the README here:
https://github.com/KDE/clazy/blob/1.11/docs/checks/README-range-loop-detach.md

Shared/alerts/DynamicEntityLayerAlertTarget.h
dtedder: Same as above

Shared/messages/MessageFeed.cpp
dtedder: Correct. No then necessary. Applied the suggested fix.

Shared/messages/MessageFeed.cpp
dtedder: Added.

Shared/messages/MessagesOverlay.cpp
dtedder: Same as above.

Shared/messages/MessagesOverlay.h
dtedder: Added = nullptr and fixed * alignments.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
JamesMBallard: I think these should have a parent here. 

However, let me take a step back and ask why you are creating graphics here. You mentioned they aren't meant for the UI. What are they meant for?

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
JamesMBallard: No, just new files added to source control should have the current year in their copyright.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
JamesMBallard: I think I answered my own question. Graphics are needed for the quadtree collision detection algorithm. Take a look at that code, or other places added graphics to it. There should be a consistent memory management technique in place.

I think there should be a QObject parent provided here, but there may be another pattern in use.

dtedder: Pushed up a new commit where 'this' is provided as the parent in the constructor. This is the case in many other instances elsewhere in the code base.

dtedder:
@ldanzinger @tanneryould 

This PR is ready for review.

Addresses #367

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+    auto* dynamicEntity = info->dynamicEntity();
+    connectEntityGraphic(dynamicEntity);
+    info->deleteLater();
+  });
+
+  // subscribe to the observation received signal (this essentially serves as the 'update' signal for the dynamic layer type)
+  connect(m_dynamicEntityLayer->dataSource(), &DynamicEntityDataSource::dynamicEntityObservationReceived, this, [this](DynamicEntityObservationInfo* observationInfo)
+  {
+    // check the entity for a connection and mark the observation as delete later
+    auto* dynamicEntity = observationInfo->observation()->dynamicEntity();
```
JamesMBallard, line 77: ditto ownership/parent here. I think we should be explicit with our expectations of who owns this. We know it won't leak since we never return object without any parent, but in cases like this we need short lifetime for DynamicEntityObservationInfo, and longer lifetime for DynamicEntity.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+/*!
+  \brief Returns the list of \l Esri::ArcGISRuntime::Geometry which are in the \a targetArea.
+
+  \note No exact intersection tests are carried out to create this list.
+ */
+QList<Geometry> DynamicEntityLayerAlertTarget::targetGeometries(const Envelope& targetArea) const
+{
+  // if the quadtree has been built, use  it to return the set of candidate geometries
+  if (m_quadtree) { return m_quadtree->candidateIntersections(targetArea); }
```
JamesMBallard, line 116: Please be consistent with indenting and style. There's nothing wrong with this style, but it's inconsistent with similar usages in this repo.
```suggestion
  if (m_quadtree) 
  { 
    return m_quadtree->candidateIntersections(targetArea); 
  }
```
Or omit the braces if you prefer, but not on the same line as the if statement. 

This comment applies throughout.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+  // otherwise, return all of the geometry in the overlay
+  QList<Geometry> geometries;
+  for (const auto* dynamicEntity : qAsConst(m_entityGraphics))
```
JamesMBallard, line 120: We found out recently that Qt has deprecated qAsConst in newer versions of Qt. We should use std::as_const as you have in other places.

https://doc.qt.io/qt-6/qttypetraits-obsolete.html#qAsConst

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+  Graphic* graphic;
+  if (m_entityGraphics.contains(dynamicEntity->entityId()))
+  {
+    // update the geometry
+    graphic = m_entityGraphics[dynamicEntity->entityId()];
+    graphic->setGeometry(dynamicEntity->geometry());
+  }
+  else
+  {
+      // if no graphic existed for the entity id, construct a new one and insert it into the graphics lookup
```
JamesMBallard, line 154:
```suggestion
    // if no graphic existed for the entity id, construct a new one and insert it into the graphics lookup
```

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+/*!
+  \internal
+
+  Connect signals etc. for a new \a dynamicEntity.
+ */
+void DynamicEntityLayerAlertTarget::connectEntityGraphic(DynamicEntity* dynamicEntity)
+{
+  // check for an existing graphic
+  Graphic* graphic;
```
JamesMBallard, line 145:
```suggestion
  Graphic* graphic = nullptr;
```

In fact, it might be cleaner to just declare this when needed in each case below.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+  // destroy the quadtree if it already existed
+  if (m_quadtree)
+  {
+    delete m_quadtree;
+    m_quadtree = nullptr;
+  }
+
+  // build a list of pointers to geoelements for every graphic in the lookup
+  QList<GeoElement*> elements;
+  for (auto* dynamicEntity : qAsConst(m_entityGraphics))
```
JamesMBallard, line 187: ditto std::as_const (see previous comments... `qAsConst` has been deprecated in newer versions of Qt).

Shared/messages/MessageFeed.cpp
```diff
@@ -17,44 +17,156 @@
...
 
-/*!
-  \brief Constructor accepting a feed \a name, a message \a type, and \a overlay and an optional \a parent.
- */
-MessageFeed::MessageFeed(const QString& name, const QString& type, MessagesOverlay* overlay, QObject* parent) :
-  QObject(parent),
-  m_feedName(name),
-  m_feedMessageType(type),
-  m_messagesOverlay(overlay)
+MessageFeed::~MessageFeed()
```
JamesMBallard, line 63: For any destructors that don't do anything, you can do `= default;` in the .cpp files.

Shared/messages/MessageFeed.cpp
```diff
@@ -17,44 +17,156 @@
...
+  // remove ownership when entities are purged
+  connect(this, &DynamicEntityDataSource::dynamicEntityPurged, this, [](DynamicEntityInfo* info)
+  {
+    // release the dynamic entity
+    auto* dynamicEntity = info->dynamicEntity();
+    dynamicEntity->setParent(nullptr);
+    dynamicEntity->deleteLater();
+
+    // mark the info as delete later so it can be cleaned up
+    info->deleteLater();
```
JamesMBallard, line 153: Nice 👍 

I think you can simplify some of this. No need to `dynamicEntity->setParent(nullptr);` if you're going to `dynamicEntity->deleteLater();` on the next line. I don't think it matters if it has a parent or not if you call `deleteLater` on it.

Shared/messages/MessageFeed.h
```diff
@@ -21,18 +21,32 @@
...
+
+}
+
 namespace Dsa {
 
+class Message;
 class MessagesOverlay;
 
-class MessageFeed : public QObject
+class MessageFeed : public Esri::ArcGISRuntime::DynamicEntityDataSource
```
JamesMBallard, line 39: Is this class now exclusively for DynamicEntity logic? If it implements the `DynamicEntityDataSource` class, it shouldn't be responsible for anything else.

Shared/qml/About.qml
```diff
@@ -105,7 +105,7 @@ Item {
 
             Text {
                 anchors.horizontalCenter: parent.horizontalCenter
-                text: "Built with ArcGIS Runtime SDK 100.9 for Qt"
+                text: "Built with ArcGIS Runtime SDK 200.3 for Qt"
```
JamesMBallard, line 108: This seems off and we may be updating it again.

Shared/alerts/AlertCondition.cpp
```diff
@@ -126,6 +133,25 @@ void AlertCondition::init(GraphicsOverlay* sourceFeed, const QString& sourceDesc
...
+  if (!sourceFeed || !target)
+    return;
+
+  m_sourceDescription = sourceDescription;
+  m_targetDescription = targetDescription;
+  const auto* dataSource = sourceFeed->dataSource();
+
+  connect(dataSource, &DynamicEntityDataSource::dynamicEntityReceived, this, [this, sourceFeed, target](DynamicEntityInfo* info)
+  {
+    auto* dynamicEntity = info->dynamicEntity();
```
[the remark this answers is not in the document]
JamesMBallard: Ok, thanks for explaining. That should be made clear somehow if it's not already.

Handheld/Handheld.pro
```diff
@@ -20,7 +20,7 @@ TEMPLATE = app
 QT += core gui opengl network positioning sensors qml quick xml
 CONFIG += c++17
 
-ARCGIS_RUNTIME_VERSION = 200.2.0
+ARCGIS_RUNTIME_VERSION = 200.4.0
```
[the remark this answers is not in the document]
dtedder: Yes. Was planning on immediately following up this branch with a new set of changes to update to 200.5 to include addressing the deprecation warnings.

Shared/DrawOrderLayerListModel.h
```diff
@@ -32,6 +32,7 @@ class DrawOrderLayerListModel : public QSortFilterProxyModel
 
 protected:
   bool lessThan(const QModelIndex& sourceLeft, const QModelIndex& sourceRight) const override;
+  bool filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const override;
```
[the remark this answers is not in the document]
dtedder: Recently updated Qt Creator. Didn't notice it reset my style settings. Had to reset to Qt Team style and disable clang again.

Shared/alerts/AlertCondition.cpp
```diff
@@ -126,6 +133,25 @@ void AlertCondition::init(GraphicsOverlay* sourceFeed, const QString& sourceDesc
     handleGraphicAt(i);
 }
 
+void AlertCondition::init(Esri::ArcGISRuntime::DynamicEntityLayer* sourceFeed, const QString& sourceDescription, AlertTarget* target, const QString& targetDescription)
```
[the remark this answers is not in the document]
dtedder: Removed.

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -402,18 +406,30 @@ bool AlertConditionsController::addAttributeEqualsAlert(const QString& condition
...
-  condition->init(sourceOverlay, sourceFeedName, target, targetValue.toString());
+
+  if (sourceGraphicsOverlay)
+  {
+    condition->init(sourceGraphicsOverlay, sourceFeedName, target, targetValue.toString());
+  }
+  else
+  {
+    condition->init(sourceDynamicEntityLayer, sourceFeedName, target, targetValue.toString());
+  }
```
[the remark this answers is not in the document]
dtedder: This approach was setup to avoid initializing anything unnecessarily but also to minimize repeating any code. Since these layers don't share an ancestor, I have switched to a templated function to avoid duplicating the same code that is necessary later in the method. Please let me know what you think.

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -605,16 +622,30 @@ void AlertConditionsController::onLayersChanged()
...
+      {
+        if (dynamicEntityLayer->loadStatus() != LoadStatus::Loaded)
+        {
+          connect(dynamicEntityLayer, &DynamicEntityLayer::doneLoading, this, &AlertConditionsController::onLayersChanged);
+        }
+        else
+        {
+          newTargetList.append(dynamicEntityLayer->name());
+          newSourceList.append(dynamicEntityLayer->name());
+          existingLayerIds.append(dynamicEntityLayer->name());
```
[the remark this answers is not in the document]
dtedder: Added.

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -1233,6 +1277,34 @@ GraphicsOverlay* AlertConditionsController::graphicsOverlayFromName(const QStrin
...
+  \brief internal
+ */
+DynamicEntityLayer* AlertConditionsController::dynamicEntityLayerFromName(const QString& layerName)
+{
+  auto* operationalLayers = ToolResourceProvider::instance()->operationalLayers();
+  if (!operationalLayers)
+    return nullptr;
+
+  const QString& layerId = m_messageFeedTypesToNames.key(layerName, layerName);
+  for (const auto& layer : *operationalLayers)
```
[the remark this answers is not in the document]
dtedder: Agreed. Will switch to pointer.

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -1233,6 +1277,34 @@ GraphicsOverlay* AlertConditionsController::graphicsOverlayFromName(const QStrin
...
+  {
+    if (!layer)
+      continue;
+    if (auto* dynamicEntityLayer = qobject_cast<DynamicEntityLayer*>(layer); dynamicEntityLayer)
+    {
+      if (dynamicEntityLayer->layerId().isEmpty())
+        continue;
+
+      qDebug(dynamicEntityLayer->layerId().toStdString().c_str());
```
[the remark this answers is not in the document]
dtedder: Removed.

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -1233,6 +1277,34 @@ GraphicsOverlay* AlertConditionsController::graphicsOverlayFromName(const QStrin
...
+
+  const QString& layerId = m_messageFeedTypesToNames.key(layerName, layerName);
+  for (const auto& layer : *operationalLayers)
+  {
+    if (!layer)
+      continue;
+    if (auto* dynamicEntityLayer = qobject_cast<DynamicEntityLayer*>(layer); dynamicEntityLayer)
+    {
+      if (dynamicEntityLayer->layerId().isEmpty())
+        continue;
```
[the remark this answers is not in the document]
dtedder: I think both of these checks in this method are unnecessary leftovers from the GraphicsOverlay method it was derived from. Removed.

Shared/alerts/DynamicEntityLayerAlertTarget.h
```diff
@@ -0,0 +1,57 @@
...
+  explicit DynamicEntityLayerAlertTarget(Esri::ArcGISRuntime::DynamicEntityLayer* dynamicEntityLayer);
+  ~DynamicEntityLayerAlertTarget();
+
+  QList<Esri::ArcGISRuntime::Geometry> targetGeometries(const Esri::ArcGISRuntime::Envelope& targetArea) const override;
+  QVariant targetValue() const override;
+
+private:
+  Esri::ArcGISRuntime::DynamicEntityLayer* m_dynamicEntityLayer = nullptr;
+  GeometryQuadtree* m_quadtree = nullptr;
+  QMap<quint64, Esri::ArcGISRuntime::Graphic*> m_entityGraphics;
```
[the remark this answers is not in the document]
dtedder: No need to sorting. Will switch to QHash

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+/*!
+  \brief Constructor taking an \l Esri::ArcGISRuntime::DynamicEntityLayer (\a dynamicEntityLayer).
+
+  Entities are added/updated/removed by the DynamicEntityDataSource of the DynamicEntityLayer.
+ */
+DynamicEntityLayerAlertTarget::DynamicEntityLayerAlertTarget(DynamicEntityLayer* dynamicEntityLayer) :
+  AlertTarget(dynamicEntityLayer),
+  m_dynamicEntityLayer(dynamicEntityLayer),
+  m_entityGraphics(QMap<quint64, Graphic*>{})
```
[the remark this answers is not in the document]
dtedder: Removed.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: Fixed throughout the file.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: Fixed.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: Fixed.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: Adjusted to declare within each case.

Shared/alerts/DynamicEntityLayerAlertTarget.cpp
dtedder: Fixed.

Shared/messages/MessageFeed.cpp
dtedder: Fixed.

Shared/messages/MessageFeed.cpp
dtedder: Wasn't sure about this one. Saw mention of it not needing to be set to null in some discussions but could not determine for sure in the Qt docs anywhere.

Removed.

Shared/qml/About.qml
dtedder: Removed the specific version number for now and replaced 'Runtime' with 'Maps SDK'. Will make a sweeping effort in an upcoming branch to update all version numbers and references to Runtime to be consistent.

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -402,18 +406,28 @@ bool AlertConditionsController::addAttributeEqualsAlert(const QString& condition
...
-  {
-    emit toolErrorOccurred(QStringLiteral("Failed to create Condition"), QString("Could not find source feed: %1").arg(sourceFeedName));
-    return false;
-  }
+  emit toolErrorOccurred(QStringLiteral("Failed to create Condition"), QString("Could not find source feed: %1").arg(sourceFeedName));
+  return false;
+}
 
+template<typename T>
+bool AlertConditionsController::addAttributeEqualsAlertBySourceLayerType(const QString& conditionName,
```
JamesMBallard, line 420: Good call, easier to understand.

PR #380, opened 2024-10-07
Added support for all types of GeoElements

### LINKED ISSUE #379: Viewshed not working when attached to a Dynamic Entity
Viewshed is not working correctly for both ways you can attach it to a Dynamic Entity:

1- Long-press on a DE (i.e. from friendly land tracks) and choose Viewshed. When this is done, the Viewshed is placed on the map where the mouse clicked, instead of on the DE as it moves. 
2- Choose "Map Element" from the Viewshed tool. Nothing happens when you click on or near a DE. 

_Note, this might work correctly for features from layers, just not DEs_

dtedder:
@JamesMBallard 
please review

Fixes #379

### DIFF
--- a/Shared/analysis/ViewshedController.cpp
+++ b/Shared/analysis/ViewshedController.cpp
@@ -176,7 +176,7 @@ void ViewshedController::onMouseClicked(QMouseEvent& event)
   if (!isActive() || !m_sceneView)
     return;
 
-  auto position = event.position();
+  const auto position = event.position();
 
   switch (m_activeMode)
   {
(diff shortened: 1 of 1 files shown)


Shared/analysis/ViewshedController.cpp
```diff
@@ -170,30 +176,84 @@ void ViewshedController::onMouseClicked(QMouseEvent& event)
   if (!isActive() || !m_sceneView)
     return;
 
+  auto position = event.position();
```
JamesMBallard, line 179:
```suggestion
  const auto position = event.position();
```

JamesMBallard (approved):
One minor thing. No need for another review.

Issue #382, opened 2024-10-08
Update version number in About screen


dtedder:
Should mention Maps SDK 200.5. 

We can keep the DSA app version number at 2.0.0

kerryrobinson:
Looks good!  👍

Issue #381, opened 2024-10-08
Use formatted names for Message Feeds


dtedder:
When creating a new condition, the drop down list of feeds uses the short names for the feeds instead of the formatted version from the config file. 
| Previous Version | Latest Version | 
| --- | --- | 
| ![Image](https://github.com/user-attachments/assets/422edb65-7241-4205-a1b2-e16ab9046ddf) | ![Image](https://github.com/user-attachments/assets/9f1a5ad0-475b-4150-b981-4a9114ea4d9e) |

kerryrobinson:
Looks good!  👍

PR #383, opened 2024-10-08
Dtedder/issues 381 feednames 382 about page

dtedder:
Few minor label changes.

@JamesMBallard 
please review

relates to #381 and #382

### DIFF
--- a/Shared/DsaResources.cpp
+++ b/Shared/DsaResources.cpp
@@ -46,6 +46,24 @@ DsaResources::~DsaResources()
 }
 
 /*!
+ * \brief Sets the string for the Maps SDK version from main.cpp
+ * \param const char* from the ARCGIS_MAPS_SDK_VERSION define
+ */
+void DsaResources::setArcGISMapsSDKVersion(const char *arcGISMapsSDKVersion)
+{
+  m_arcGISMapsSDKVersion = QString(arcGISMapsSDKVersion);
+}
+
+/*!
+ * \brief Returns the version string of the Maps SDK that is used by the project
+ * \return
+ */
+QString DsaResources::arcGISMapsSDKVersion() const
+{
+  return m_arcGISMapsSDKVersion;
+}
+
+/*!
   \brief Returns the URL to the "2D" icon.
  */
 QUrl DsaResources::icon2d() const
--- a/Shared/DsaResources.cpp
+++ b/Shared/DsaResources.cpp
@@ -46,17 +46,15 @@ DsaResources::~DsaResources()
 }
 
 /*!
- * \brief Sets the string for the Maps SDK version from main.cpp
- * \param const char* from the ARCGIS_MAPS_SDK_VERSION define
+ * \brief Sets the string for the Maps SDK version from the #define for ARCGIS_MAPS_SDK_VERSION in main.cpp
  */
-void DsaResources::setArcGISMapsSDKVersion(const char *arcGISMapsSDKVersion)
+void DsaResources::setArcGISMapsSDKVersion(const char* arcGISMapsSDKVersion)
 {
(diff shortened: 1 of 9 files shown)


Shared/qml/About.qml
```diff
@@ -105,7 +105,7 @@ Item {
 
             Text {
                 anchors.horizontalCenter: parent.horizontalCenter
-                text: "Built with ArcGIS Maps SDK for Qt"
+                text: "Built with ArcGIS Maps SDK 200.5 for Qt"
```
JamesMBallard, line 108: How can we keep track of this when we change the version?

dtedder: Pushed another commit that will ultimately derive the values from the version number listed in the .pro files.

Shared/DsaResources.cpp
```diff
@@ -45,6 +45,24 @@ DsaResources::~DsaResources()
 { 
 }
 
+/*!
+ * \brief Sets the string for the Maps SDK version from main.cpp
+ * \param const char* from the ARCGIS_MAPS_SDK_VERSION define
+ */
+void DsaResources::setArcGISMapsSDKVersion(const char *arcGISMapsSDKVersion)
```
JamesMBallard, line 52:
```suggestion
void DsaResources::setArcGISMapsSDKVersion(const char* arcGISMapsSDKVersion)
```

Shared/DsaResources.cpp
```diff
@@ -45,6 +45,24 @@ DsaResources::~DsaResources()
...
+{
+  m_arcGISMapsSDKVersion = QString(arcGISMapsSDKVersion);
+}
+
+/*!
+ * \brief Returns the version string of the Maps SDK that is used by the project
+ * \return
```
JamesMBallard, line 59: This looks incomplete, but I don't know if we actually run qdoc on this project anymore.
@ldanzinger do you know about that?

Shared/DsaResources.h
```diff
@@ -27,6 +27,7 @@ class DsaResources : public QObject
 {
   Q_OBJECT
 
+  Q_PROPERTY(QString arcGISMapsSDKVersion READ arcGISMapsSDKVersion CONSTANT)
```
JamesMBallard, line 30: Since technically this can be empty to begin with, it should have a NOTIFY and a corresponding signal along with it when it changes.

Shared/DsaResources.cpp
dtedder: moved pointer to type

Shared/DsaResources.h
dtedder: removed CONSTANT. added NOTIFY and a signal for it.

Shared/DsaResources.cpp
dtedder: I removed the return statement here. Perhaps it's good enough for now to just include a *brief* for methods?

JamesMBallard: Probably fine for now.

Shared/DsaResources.cpp
```diff
@@ -45,6 +45,22 @@ DsaResources::~DsaResources()
 { 
 }
 
+/*!
+ * \brief Sets the string for the Maps SDK version from the #define for ARCGIS_MAPS_SDK_VERSION in main.cpp
+ */
+void DsaResources::setArcGISMapsSDKVersion(const char* arcGISMapsSDKVersion)
+{
+  m_arcGISMapsSDKVersion = QString(arcGISMapsSDKVersion);
+}
```
JamesMBallard, line 54:
```suggestion
  m_arcGISMapsSDKVersion = QString(arcGISMapsSDKVersion);
  emit arcGISMapsSDKVersionChanged();
}
```

JamesMBallard (approved):
One last thing. No need for another review.

Issue #384, opened 2024-10-08
Update UI for Conditions and Observations Report panels


dtedder:
The `Conditions` and `Observations Report` panels UIs needs to be updated. The text on the Back and Next buttons are not visible (instead you see ellipses).  Also, hitting the X at the bottom ('Cancel')  does not cancel the panel (hitting the X in the upper right-hand corner does). 
![image](https://github.com/Esri/dynamic-situational-awareness-qt/assets/3383476/ccfba940-e582-445c-a2a1-d1f257a7d3f9)

This might be happening in other places too, need to log any UI issues here.

dtedder:
@kerryrobinson These look ok? I tested on iOS actual device and things appear to scale fairly well.

![image](https://github.com/user-attachments/assets/4ec8b431-3876-4c81-9bc6-e28f2e71b9a1)![image](https://github.com/user-attachments/assets/6185eb00-c54d-4244-a41b-b7571bbc23c2)![image](https://github.com/user-attachments/assets/81537b86-1e6d-497b-9ee7-7c86a60cf279)![image](https://github.com/user-attachments/assets/f625f8af-541d-40b9-bf50-954a7eeab750)

kerryrobinson:
Yes that looks much better and at least works for now, since the text is readable.  It would be great if the text was larger (but didn't get cut off). The buttons could also be square instead of rounded which might accommodate this. If this doesn't end up being a simple fix, we can come back to this later and look at larger UI updates to the app.

kerryrobinson:
This looks much better now.  👍

PR #385, opened 2024-10-09
Cleanup Alerts/Observation wizard buttons

dtedder:
@JamesMBallard
please review.

relates to #384 

- added a property to store the margin factor needed around the wizard button row

- wrapped the buttons in a RowLayout to manage their alignment and spacing
- implemented same logic for both wizard panels
	- hiding forward and back buttons but making them take up the same space so controls don't shift around
	- keeping both create and cancel visible but fading create button when it isn't usable

### DIFF
--- a/Shared/qml/AlertConditionsWizard.qml
+++ b/Shared/qml/AlertConditionsWizard.qml
@@ -222,7 +222,11 @@ Rectangle {
             height: nextButton.height
             width: nextButton.width
             text: "Back"
-            font.pixelSize: DsaStyles.toolFontPixelSize * scaleFactor * 1.75
+            leftPadding: 0
+            rightPadding: 0
+            topPadding: 0
+            bottomPadding: 0
+            font.pixelSize: DsaStyles.toolFontPixelSize * scaleFactor * 1.5
 
             onClicked: conditionFrame.decrementCurrentIndex();
         }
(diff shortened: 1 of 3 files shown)


Shared/qml/AlertConditionsWizard.qml
```diff
@@ -26,6 +27,7 @@ Rectangle {
 
     property bool readyToAdd: conditionFrame.currentItem == reviewPage
     property real scaleFactor: (Screen.logicalPixelDensity * 25.4) / (Qt.platform.os === "windows" || Qt.platform.os === "linux" ? 96 : 72)
+    property real wizardButtonsFactoredMargin: 4 * scaleFactor
```
JamesMBallard, line 30:
```suggestion
    readonly property real wizardButtonsFactoredMargin: 4 * scaleFactor
```

Shared/qml/ObservationReportTool.qml
```diff
@@ -29,6 +30,7 @@ DsaPanel {
     property alias pickMode: toolController.pickMode
     property bool  toolActive: toolController.active
     property bool readyToAdd: reportFrame.currentItem == reviewPage
+    property real wizardButtonsFactoredMargin: 4 * scaleFactor
```
JamesMBallard, line 33:
```suggestion
    readonly property real wizardButtonsFactoredMargin: 4 * scaleFactor
```

JamesMBallard (approved):
A few suggestions, but no need for another review. I admit it's difficult to review QML code related to UI changes.

Issue #392, opened 2024-10-17
LOS tool not working


dtedder:
Line of Sight is not working in two places:

1- Using the Line of Sight tool from the Analysis toolbar. I can select a points layer as the source, but it does not show any results, AND I do not see any results in the "Analyst List". 

~2- When long-pressing on a friendly land track, if I select LIne of Sight, no LOS is displayed.~ (this was never supported)

~In this case, I can see the LOS in the "Analyst List", it's just not drawing on the map.~ 

![Image](https://github.com/user-attachments/assets/7800465e-b30e-46ee-8a60-ce40e9d62144)

PR #393, opened 2024-10-17
Implementing DynamicEntity layer with LOS tools

dtedder:
@JamesMBallard  please review

related #392 

- fetched the DynamicEntity from the DynamicEntityObservation returned on long press/context menu
- moved updateLayerNames from a lambda to a member function
- fixed an issue where layers were getting skipped due to a check returning early from missing curly braces missed during the 200.5 migration to QFuture
- implemented layer support for DynamicEntity types
- factored out code that was common to both FeatureLayer and DynamicEntityLayer
- using generic Layer for m_overlays list to allow for both FeatureLayer and DynamicEntityLayer

### DIFF
--- a/Shared/ContextMenuController.cpp
+++ b/Shared/ContextMenuController.cpp
@@ -20,6 +20,8 @@
 #include "ContextMenuController.h"
 
 // C++ API headers
+#include "DynamicEntity.h"
+#include "DynamicEntityObservation.h"
 #include "Graphic.h"
 #include "GraphicsOverlay.h"
 #include "IdentifyGraphicsOverlayResult.h"
@@ -480,7 +482,14 @@ void ContextMenuController::selectOption(const QString& option)
           if (!geoElement || geoElement->geometry().geometryType() != GeometryType::Point)
             continue;
 
-          lineOfSightTool->lineOfSightFromLocationToGeoElement(geoElement);
+          // identify results of type observation require the dynamic entity they belong to
+          if (auto* dynamicEntityObservation = dynamic_cast<DynamicEntityObservation*>(geoElement); dynamicEntityObservation)
+          {
+            auto* dynamicEntityGeoElement = static_cast<GeoElement*>(dynamicEntityObservation->dynamicEntity());
+            lineOfSightTool->lineOfSightFromLocationToGeoElement(dynamicEntityGeoElement);
+          }
+          else
+            lineOfSightTool->lineOfSightFromLocationToGeoElement(geoElement);
         }
       }
     };
--- a/Shared/ContextMenuController.cpp
+++ b/Shared/ContextMenuController.cpp
@@ -489,7 +489,9 @@ void ContextMenuController::selectOption(const QString& option)
             lineOfSightTool->lineOfSightFromLocationToGeoElement(dynamicEntityGeoElement);
           }
           else
+          {
             lineOfSightTool->lineOfSightFromLocationToGeoElement(geoElement);
+          }
         }
       }
     };
--- a/Shared/analysis/LineOfSightController.cpp
+++ b/Shared/analysis/LineOfSightController.cpp
@@ -23,6 +23,7 @@
 #include "AnalysisListModel.h"
 #include "AnalysisOverlay.h"
 #include "AnalysisOverlayListModel.h"
+#include "DynamicEntity.h"
 #include "Error.h"
 #include "Feature.h"
 #include "FeatureIterator.h"
@@ -46,6 +47,7 @@
 #include "FeatureQueryResultManager.h"
 #include "LocationController.h"
 #include "LocationDisplay3d.h"
+#include "MessagesOverlay.h"
 #include "ToolManager.h"
 #include "ToolResourceProvider.h"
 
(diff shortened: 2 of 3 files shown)


Shared/analysis/LineOfSightController.h
```diff
@@ -81,10 +82,13 @@ public slots:
   void cancelTask();
   void getLocationGeoElement();
   void setVisibleByCount(int visibleByCount);
+  void updateLayerNames();
+  bool resetAnalysis(size_t featureCount);
```
JamesMBallard, line 86: Is size_t the best choice? Perhaps qsizetype or something similar? size_t seems ok.

Shared/analysis/LineOfSightController.cpp
```diff
@@ -391,98 +403,129 @@ void LineOfSightController::selectOverlayIndex(int selectOverlayIndex)
...
+  setVisibleByCount(0);
 
-          if (lineOfSight->targetVisibility() == LineOfSightTargetVisibility::Visible)
-            visibleCount += 1;
-        }
+  // clear the QObject used as a parent for Line of Sight results
+  if (m_lineOfSightParent)
+  {
+    delete m_lineOfSightParent;
+    m_lineOfSightParent = nullptr;
```
JamesMBallard, line 494: I assume the optimizer will remove this since it's overwritten on the next line. I think you can remove it.
```suggestion
```

Another option is to use a `std::unique_ptr<QObject>` as the parent. Then you just need to do a `m_lineOfSightParent = std::make_unique<QObject>();` and that does all this in a clearer way.

Shared/analysis/LineOfSightController.cpp
```diff
@@ -391,98 +403,129 @@ void LineOfSightController::selectOverlayIndex(int selectOverlayIndex)
...
-    return true;
-  }).then(this, [this](bool success)
+  m_visibleByConnections.append(connect(lineOfSight, &GeoElementLineOfSight::targetVisibilityChanged, this, [this]()
   {
-    if (!success)
-      emit selectOverlayFailed();
-  });
+    int visibleCount = 0;
+    const auto* losList = m_lineOfSightOverlay->analyses();
+    for (auto losIt = losList->begin(); losIt < losList->end(); losIt++)
```
JamesMBallard, line 509:
```suggestion
    for (auto losIt = std::being(losList); losIt != std::end(losList); ++losIt)
```

I don't think using less than comparisons with iterators is typical.

Shared/analysis/LineOfSightController.cpp
```diff
@@ -391,98 +403,129 @@ void LineOfSightController::selectOverlayIndex(int selectOverlayIndex)
...
+    {
+      const auto* analysis = *losIt;
+      if (!analysis)
+        continue;
+
+      const auto* lineOfSight = qobject_cast<const GeoElementLineOfSight*>(analysis);
+      if (!lineOfSight)
+        continue;
```
JamesMBallard, line 517: Can you combine these? qobject_cast should be able to handle null objects. Double check.

Shared/ContextMenuController.cpp
```diff
@@ -480,7 +482,14 @@ void ContextMenuController::selectOption(const QString& option)
...
 
-          lineOfSightTool->lineOfSightFromLocationToGeoElement(geoElement);
+          // identify results of type observation require the dynamic entity they belong to
+          if (auto* dynamicEntityObservation = dynamic_cast<DynamicEntityObservation*>(geoElement); dynamicEntityObservation)
+          {
+            auto* dynamicEntityGeoElement = static_cast<GeoElement*>(dynamicEntityObservation->dynamicEntity());
+            lineOfSightTool->lineOfSightFromLocationToGeoElement(dynamicEntityGeoElement);
+          }
+          else
+            lineOfSightTool->lineOfSightFromLocationToGeoElement(geoElement);
```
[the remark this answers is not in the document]
dtedder: Agreed. If any in the if statement require braces, they should all get braces.

Shared/analysis/LineOfSightController.cpp
```diff
@@ -99,6 +101,64 @@ void LineOfSightController::setVisibleByCount(int visibleByCount)
...
+  // remove any previously tracked overlay names from the list
+  m_overlays.clear();
+
+  // check the scene's operational layers
+  QStringList overlayNames;
+  auto* operationalLayers = ToolResourceProvider::instance()->operationalLayers();
+  if (operationalLayers)
+  {
+    // loop through all the layers in the operationalLayers property
+    for (auto layerIt = operationalLayers->begin(); layerIt < operationalLayers->end(); layerIt++)
```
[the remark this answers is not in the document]
dtedder: actually meant to use ranged for loop here. will switch to that and remove the need for another variable declaration

Shared/analysis/LineOfSightController.h
dtedder: will switch to qsizetype since that is the type used by the message list count property. The other source, feature count, returns type qint64.

Shared/analysis/LineOfSightController.cpp
dtedder: Switching to ranged loop here as well.

Shared/analysis/LineOfSightController.cpp
dtedder: These will be combined with the switch to range loop.

JamesMBallard (approved):

Issue #394, opened 2024-10-22
When restarting an app that has a saved Condition, that condition does not start


dtedder:
Steps to reproduce:
- create a condition
- close the app
- open the app. the condition is in the list but is not running.

Background: 
I created a condition, it worked in the app (but then the app crashed when I tried to turn off the Distress condition). It saved the condition to the config file, so when I started the app again I can see it in the list, but no alerts are being triggered, whether I check or uncheck that condition. 

NOTE: May be fixed in the latest commit

kerryrobinson:
@kerryrobinson  tests again, with both position reports and CoT  feeds

kerryrobinson:
I'm still seeing this, will demonstrate when we meet.

kerryrobinson:
Rested with the fix in the latest branch and this is looking good. This screenshot is after creating several conditions, closing the app and reopening it.  All names are as expected and alerts showing as expected:
![Image](https://github.com/user-attachments/assets/edfc9d97-f5dc-49f3-8ec1-79c7cabfab37)

kerryrobinson:
Closing this issue as fixed

PR #395, opened 2024-10-22
added a check method to prevent duplicate alerts from being added on startup

dtedder:
@JamesMBallard 
please review

The introduction of QFuture caused a similar loading issue when a unique feature from within a layer used as the target is selected for the alert. When the query goes out to the table, subsequent calls to this method can occur and cause multiple alerts to be created for the same alert as other properties change within the application's configuration.

#394

### DIFF
--- a/Shared/alerts/AlertConditionsController.cpp
+++ b/Shared/alerts/AlertConditionsController.cpp
@@ -296,6 +296,7 @@ QFuture<bool> AlertConditionsController::addWithinDistanceAlertBySourceLayerType
   // make sure the condition has not already been added
   if (conditionAlreadyAdded(conditionName))
     return QtFuture::makeReadyFuture(false);
+
   auto* condition = new WithinDistanceAlertCondition(level, conditionName, distance, this);
   connect(condition, &WithinDistanceAlertCondition::newConditionData, this, &AlertConditionsController::handleNewAlertConditionData);
   condition->init(alertSourceLayer, sourceFeedName, target, targetDescription);
@@ -374,6 +375,7 @@ QFuture<bool> AlertConditionsController::addWithinAreaAlertBySourceLayerType(con
   // make sure the condition has not already been added
   if (conditionAlreadyAdded(conditionName))
     return QtFuture::makeReadyFuture(false);
+
   auto* condition = new WithinAreaAlertCondition(level, conditionName, this);
   connect(condition, &WithinAreaAlertCondition::newConditionData, this, &AlertConditionsController::handleNewAlertConditionData);
   condition->init(alertSourceLayer, sourceFeedName, target, targetDescription);
@@ -438,6 +440,7 @@ QFuture<bool> AlertConditionsController::addAttributeEqualsAlertBySourceLayerTyp
   // make sure the condition has not already been added
   if (conditionAlreadyAdded(conditionName))
     return QtFuture::makeReadyFuture(false);
+
   AlertTarget* target = new FixedValueAlertTarget(targetValue, this);
   AttributeEqualsAlertCondition* condition = new AttributeEqualsAlertCondition(level, conditionName, attributeName, this);
   connect(condition, &AttributeEqualsAlertCondition::newConditionData, this, &AlertConditionsController::handleNewAlertConditionData);
--- a/Shared/alerts/AlertConditionsController.cpp
+++ b/Shared/alerts/AlertConditionsController.cpp
@@ -293,6 +293,9 @@ QFuture<bool> AlertConditionsController::addWithinDistanceAlertBySourceLayerType
                                                                                  const QString& targetDescription,
                                                                                  T* alertSourceLayer)
 {
+  // make sure the condition has not already been added
+  if (conditionAlreadyAdded(conditionName))
+    return QtFuture::makeReadyFuture(false);
   auto* condition = new WithinDistanceAlertCondition(level, conditionName, distance, this);
   connect(condition, &WithinDistanceAlertCondition::newConditionData, this, &AlertConditionsController::handleNewAlertConditionData);
   condition->init(alertSourceLayer, sourceFeedName, target, targetDescription);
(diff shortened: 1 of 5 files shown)


Shared/alerts/AlertConditionsController.cpp
```diff
@@ -293,6 +293,9 @@ QFuture<bool> AlertConditionsController::addWithinDistanceAlertBySourceLayerType
                                                                                  const QString& targetDescription,
                                                                                  T* alertSourceLayer)
 {
+  // make sure the condition has not already been added
+  if (conditionAlreadyAdded(conditionName))
+    return QtFuture::makeReadyFuture(false);
   auto* condition = new WithinDistanceAlertCondition(level, conditionName, distance, this);
```
JamesMBallard, line 299:
```suggestion
    return QtFuture::makeReadyFuture(false);

  auto* condition = new WithinDistanceAlertCondition(level, conditionName, distance, this);
```

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -368,6 +371,9 @@ QFuture<bool> AlertConditionsController::addWithinAreaAlertBySourceLayerType(con
                                                                              const QString& targetDescription,
                                                                              T* alertSourceLayer)
 {
+  // make sure the condition has not already been added
+  if (conditionAlreadyAdded(conditionName))
+    return QtFuture::makeReadyFuture(false);
   auto* condition = new WithinAreaAlertCondition(level, conditionName, this);
```
JamesMBallard, line 377:
```suggestion
    return QtFuture::makeReadyFuture(false);

  auto* condition = new WithinAreaAlertCondition(level, conditionName, this);
```

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -429,6 +435,9 @@ QFuture<bool> AlertConditionsController::addAttributeEqualsAlertBySourceLayerTyp
                                                                                   const QVariant& targetValue,
                                                                                   T* alertSourceLayer)
 {
+  // make sure the condition has not already been added
+  if (conditionAlreadyAdded(conditionName))
+    return QtFuture::makeReadyFuture(false);
```
JamesMBallard, line 440:
```suggestion
  if (conditionAlreadyAdded(conditionName))

    return QtFuture::makeReadyFuture(false);
```

JamesMBallard (commented):
Are conditions unique to their name?

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -429,6 +435,9 @@ QFuture<bool> AlertConditionsController::addAttributeEqualsAlertBySourceLayerTyp
...
   AlertTarget* target = new FixedValueAlertTarget(targetValue, this);
```
JamesMBallard, line 441:
```suggestion
  if (conditionAlreadyAdded(conditionName))
    return QtFuture::makeReadyFuture(false);

  AlertTarget* target = new FixedValueAlertTarget(targetValue, this);
```

Shared/qml/AlertConditionsTool.qml
```diff
@@ -244,6 +248,8 @@ DsaPanel {
             width: parent.width
             spacing: 10 * scaleFactor
             leftPadding: 10 * scaleFactor
+            property bool nameAlreadyInUseMobile: !toolController.conditionAlreadyAdded(editConditionMobileName.text)
+            property bool validMobile: editConditionMobileName.length > 0 && nameAlreadyInUseMobile
```
JamesMBallard, line 252: This is confusing to me. It is valid if it has some length and is already in use?

Is `nameAlreadyInUseMobile: !toolController.conditionAlreadyAdded...` correct?

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -1058,6 +1070,17 @@ void AlertConditionsController::addStoredConditions()
...
+bool AlertConditionsController::conditionAlreadyAdded(const QString& conditionName)
+{
+  for (auto i = 0; i < m_conditions->rowCount(); i++)
+  {
+    const auto* condition = m_conditions->conditionAt(i);
+    if (condition->name() == conditionName)
+      return true;
+  }
+  return false;
+}
```
JamesMBallard, line 1082: Is it possible to use a range loop? If not, this is fine.

Shared/qml/AlertConditionsTool.qml
dtedder: Yep, this logic was named wrong/backwards. I started it on the main naming page for the alert and somehow managed to transpose the logic here. Was made to work properly, just used backwards for its name.

Shared/alerts/AlertConditionsController.cpp
dtedder: This class itself (subclass of QAbstractItemModel) does not directly support ranged loop. But we can add the following to the header to expose the underlying QList<AlertConditions*> which is the root object really. Not many examples in the codebase where we define the body in the header file. This seems simple enough to just put in the header though.

```
QList<AlertCondition*>::iterator begin() { return m_conditions.begin(); }
QList<AlertCondition*>::iterator end() { return m_conditions.end(); }
const QList<AlertCondition*>::const_iterator const_begin() { return m_conditions.constBegin(); }
const QList<AlertCondition*>::const_iterator const_end() { return m_conditions.constEnd(); }
```

Shared/qml/AlertConditionsTool.qml
dtedder: added commit to fix logic
9dbf34e0e5ddb97538963ac073c2bd74dc0f7741

Shared/alerts/AlertConditionsController.cpp
JamesMBallard: Let's see if this becomes a persistent issue. Short of that I don't think it's worth solving in isolation for this one case.

JamesMBallard (approved):

Issue #368, opened 2023-10-25
Update async code to use QFuture tasks instead of TaskWatcher


ldanzinger:
Asynchronous methods that returned a TaskWatcher now have new versions that return a QFuture. The new syntax allows for many new benefits, such as chaining with `.then`. With 200.3, the TaskWatcher methods will be deprecated, so we need to start migrating app code to use the new QFuture versions

kerryrobinson:
All warnings should be fixed now, in review.

kerryrobinson:
Kerry to verify that there are no deprecation warnings.

JamesMBallard:
This is more of a build thing. I trust @dtedder to be able to determine if there are no more deprecation warnings with the build. I don't think Kerry needs to verify this one. Just my 2 cents.

kerryrobinson:
Agreed.  FYI I did get some TaskWatcher warnings, but all relating to the Qt SDK deployment, not the DSA code, i.e.:
![Image](https://github.com/user-attachments/assets/6835cd7d-2c59-48a8-9032-f7121268ae76)

If this is expected, we can close this issue.

JamesMBallard:
You should not be seeing those anymore, but it looks like indirect inclusion on TaskWatcher.h and it's not used anywhere. @dtedder can you check to see how TaskWatcher is getting included? It should be forward declared everywhere in our SDK headers, and if it's unused then TaskWatcher.h should never be included.

It's possible there are no issues here, but I am curious to figure out why these are popping up. We wouldn't want our users to see a flood of warnings like this for classes they aren't using.

dtedder:
Maybe here?
https://github.com/Esri/dynamic-situational-awareness-qt/blob/01c0e13df0a4ed1a7e90b4d7ec80e21741a123ff/Shared/pch.hpp#L37

The only other references I can find are from the Toolkit which still includes it in a few files.

JamesMBallard:
Yes that's it. That should be removed now. The toolkit has been updated to no longer use TaskWatcher. @dtedder can you make that tweak to pch.hpp?

PR #396, opened 2024-10-30
Remove remaining include for TaskWatcher from DSA

dtedder:
@JamesMBallard 

patch to remove the include from the precompiled header

#368


JamesMBallard (approved):
👍

Issue #375, opened 2024-08-14
Add ability to download data from Online portal item


kerryrobinson:
To make it easier for users to get the demo data, especially on mobile platforms, let's add an option to download data (instead of sideloading it). It could work similar to the download options we already have in our samples app.  This should work both for the demo data we provide in ArcGIS Online (the Monterey data) and for any custom data that has been created (in other parts of the world, for instance). 

_This issue is specific to the ability to download the default Monterey demo data for the app_

Requirements: 
- New tool (download icon) to "Download Data"  - Could be in DSA Settings
  - If user opens the app and other is currently no data in the DSA folder, user should be directed too the DSA Packages tool so they can download the data. 
- Options for where to find the data:
  - "DSA Demo Data" - this is a well-known Portal item that contains a zip file. The user doesn't need to know the exact portal item URL, it just goes straight there and downloads (like how it works in the Samples app)
  - ~"Custom Data"~ (this is part of the follow-on issue https://github.com/Esri/dynamic-situational-awareness-qt/issues/402)
- Data is unzipped into the DSA data folder, according to the platform it's deployed on
- If data already exists, ask the user if they want to wipe it first, otherwise cancel/abort!

Questions:
- how to handle overwriting?  how to know there is existing data there? 
  - can possibly give the user the option to wipe the existing data and replace. 
- should we allow the user to manage two sets of data?  
  - ultimately this would be great, but for now let's just focus on downloading the data.  we should create an issue to capture how it could work managing more than one dataset, etc. 
- How to handle if the data isn't in the correct folders, etc?  
  - Just unzip anyway. The app will open and let the user know that it can't find. 

Definition of Done:
- [x] Design discussed and agreed
- [x] Create new issue for the ability to specify a URL/path to custom data to download (but the user can manually copy the data into the folder structure) https://github.com/Esri/dynamic-situational-awareness-qt/issues/402
- [x] Implementation done
- [ ] ~Doc updated~  (covered in https://github.com/Esri/dynamic-situational-awareness-qt/issues/370)
- [x] Tested and verified

kerryrobinson:
Let's also see if this tool could accommodate pointing to a user-specified zip file of data.

kerryrobinson:
Discussed:

- "DSA Package" zip file
  - everything must be according to the prescribed folder structure
  - all packages will be put into a subfolder of DSA, with a descriptive name ("SanDiego")
[...]

PR #399, opened 2025-01-13
Dtedder/issue 375 add package manager

dtedder:
Implementation for the Configurations feature.

@JamesMBallard 
#375


Shared/configurations/ConfigurationListModel.cpp
```diff
@@ -0,0 +1,142 @@
...
+namespace Dsa {
+
+ConfigurationListModel::ConfigurationListModel(QObject* parent):
+  QAbstractListModel(parent)
+{
+}
+
+ConfigurationListModel::~ConfigurationListModel()
+{
+}
```
JamesMBallard, line 30:
```suggestion
ConfigurationListModel::~ConfigurationListModel() = default;
```

Throughout for any empty constructors or destructors.

Shared/configurations/ConfigurationController.h
```diff
@@ -0,0 +1,89 @@
...
+// Qt headers
+#include <QDir>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+
+// DSA headers
+#include "AbstractTool.h"
+#include "ZipHelper.h"
+
+Q_MOC_INCLUDE("qabstractitemmodel.h")
```
JamesMBallard, line 29:
```suggestion
Q_MOC_INCLUDE("QAbstractListModel")
```

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+                                        configurationObject["url"].toString(),
+                                        selected,
+                                        selected && downloaded,
+                                        downloaded ? 100 : 0);
+        }
+      }
+    }
+  }
+  emit configurationsChanged();
+}
```
JamesMBallard, line 371: This method is doing a lot of work. Can it be:
1. Decomposed into smaller pieces?
2. Refactored to reduce the nesting level, possibly with early returns?

A quick count looks like there are 8 nesting levels to this.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+  connect(m_zipHelper, &ZipHelper::extractError, this, &ConfigurationController::extractError);
+
+  // extract the file to the associated configuration folder
+  auto activeDownloadIndex = m_configurationListModel->index(m_activeDownloadIndex);
+  auto activeDownloadName = m_configurationListModel->data(activeDownloadIndex, ConfigurationListModel::Roles::Name).toString();
+  QDir dirConfigurations{DsaUtility::configurationsDirectoryPath()};
+  QDir dirActiveDownloadConfiguration{dirConfigurations.filePath(activeDownloadName)};
+  if (!dirActiveDownloadConfiguration.exists())
+    dirConfigurations.mkdir(activeDownloadName);
+  m_zipHelper->extractAll(dirActiveDownloadConfiguration.absolutePath());
```
JamesMBallard, line 270:
```suggestion
  if (!dirActiveDownloadConfiguration.exists())
    dirConfigurations.mkdir(activeDownloadName);

  m_zipHelper->extractAll(dirActiveDownloadConfiguration.absolutePath());
```

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+  {
+    QFile fileFinish{m_downloadFileName};
+    if (!fileFinish.open(QIODevice::Append))
+    {
+      emit toolErrorOccurred("Unable to complete download", "Error Downloading");
+      return;
+    }
+
+    fileFinish.write(m_networkReply->readAll());
+    fileFinish.flush();
```
JamesMBallard, line 252: Is this needed? Presumably QFile will flush all data on destruction, which is guaranteed at the end of this block.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+void ConfigurationController::finished()
+{
+  // turn off the downloading state and clean up any aborted download files remaining
+  m_downloading = false;
+  if (m_aborted)
+  {
+    QFile fileAbort(m_downloadFileName);
+    if (fileAbort.exists())
+      fileAbort.remove();
+    m_aborted = false;
```
JamesMBallard, line 237: There are a few of these. I won't comment on them all to keep the review comments to a minimum, but please add a line after if statement conditions.
```suggestion
    if (fileAbort.exists())
      fileAbort.remove();

    m_aborted = false;
```

IMO it can hide conditional logic when someone new is reading through the code. Alternately you can always use braces for single-statement condition blocks.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+QAbstractListModel* ConfigurationController::configurations() const
+{
+  return m_configurationListModel;
+}
+
+void ConfigurationController::downloadProgress(quint64 bytesReceived, quint64 bytesTotal)
+{
+  setPercentComplete(bytesReceived * 1.0 / bytesTotal * 1.0 * 100);
```
JamesMBallard, line 211: Guard against division by zero. Seems unlikely, but good to be safe.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+QString ConfigurationController::toolName() const
+{
+  return QStringLiteral("configurations");
+}
+
+void ConfigurationController::setProperties(const QVariantMap& properties)
+{
+  Q_UNUSED(properties);
+}
```
JamesMBallard, line 197: This is a big PR so I'm having difficulties navigating all the code (not your fault, just lots of zlib stuff). But I cannot tell if this is an interface or something. What is the purpose of this method?

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+      }
+    }
+  }
+}
+
+void ConfigurationController::remove(int index)
+{
+  // make extracted packages removeable from device
+  Q_UNUSED(index);
+}
```
JamesMBallard, line 149: Same question... Why have this?

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+void ConfigurationController::cancel(int index)
+{
+  Q_UNUSED(index); // TODO: supporting multiple downloads at a time will require additional management
+  if (m_networkReply)
+  {
+    if (m_networkReply != nullptr)
+    {
+      if (m_networkReply->isOpen())
```
JamesMBallard, line 132: `if (m_networkReply && m_networkReply->isOpen())` should cover all these checks and reduce the nesting level.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+  {
+    // get the total bytes that will be downloaded
+    bool convertedOk = false;
+    int bytesToDownload = 0;
+    auto contentLengthVariant = headReply->header(QNetworkRequest::ContentLengthHeader);
+    if (contentLengthVariant.isValid())
+      bytesToDownload = contentLengthVariant.toInt(&convertedOk);
+
+    // clean up the head reply
+    headReply->deleteLater();
```
JamesMBallard, line 97: I recommend using autoDeleteReplies feature. That should eliminate the need to manually `deleteLater` any reply. We were able to clean up some of our code as well with that.

https://doc.qt.io/qt-6.5/qnetworkaccessmanager.html#setAutoDeleteReplies

Just be careful you're not holding onto replies after `finished` happens.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,373 @@
...
+ConfigurationController::ConfigurationController(QObject* parent /* = nullptr */):
+  AbstractTool(parent),
+  m_configurationListModel(new ConfigurationListModel{this})
+{
+  connect(m_configurationListModel, &ConfigurationListModel::dataChanged, this, [this]() { emit configurationsChanged(); });
+
+  ToolManager::instance().addTool(this);
+
+  // set the downloads folder based on the system type
+#if defined Q_OS_IOS || defined Q_OS_ANDROID
```
JamesMBallard, line 51: QtGlobal should be included above to ensure the platform defines are available. It is likely being transitively included already, but it's best to be explicit.

Shared/configurations/Configuration.h
```diff
@@ -0,0 +1,65 @@
...
+
+// Qt headers
+#include <QString>
+
+namespace Dsa {
+
+class Configuration
+{
+public:
+  explicit Configuration(const QString& name,
```
JamesMBallard, line 28: remove `explicit` since that's only for single-arg constructors. Reindent as well.

Shared/configurations/Configuration.h
```diff
@@ -0,0 +1,65 @@
...
+                         const QString& url,
+                         bool selected,
+                         bool loaded,
+                         int percentDownloaded);
+  Configuration(Configuration&) = default;
+  Configuration(const Configuration&) = default;
+  Configuration(Configuration&&) = default;
+  Configuration& operator=(const Configuration&) = default;
+  Configuration& operator=(Configuration&&) = default;
```
JamesMBallard, line 37: 👍 cool

Shared/configurations/Configuration.h
```diff
@@ -0,0 +1,65 @@
...
+
+  int percentDownloaded() const;
+  void setPercentDownloaded(int percentDownloaded);
+
+private:
+  QString m_name;
+  QString m_url;
+  bool m_selected;
+  bool m_loaded;
+  int m_percentDownloaded;
```
JamesMBallard, line 60: Although it's not needed now since you only have a single, parameterized constructor, provide defaults here for the bools and int. If someone comes along later and adds a new constructor that does not require these, they will be uninitialized.
```suggestion
  bool m_selected = false;
  bool m_loaded = false;
  int m_percentDownloaded = 0;
```

Shared/DsaUtility.cpp
```diff
@@ -41,23 +50,95 @@ namespace Dsa {
...
-QString DsaUtility::dataPath()
+QString DsaUtility::configurationsDirectoryPath()
 {
-  QDir dataDir;
-#ifdef Q_OS_ANDROID
-  dataDir = QDir(QStandardPaths::writableLocation(QStandardPaths::HomeLocation));
-#elif defined Q_OS_IOS
-  dataDir = QDir(QStandardPaths::writableLocation(QStandardPaths::AppDataLocation));
+  // return the cached path to the configurations folder
+  static QString path;
```
JamesMBallard, line 56: Is there any chance any of these cached values can be changed? Your new caching logic will introduce problems if that's the case. I doubt that's the case, however.

Also, for these I would consider `static QString configurationsDirectoryPath;` or something more descriptive than just `path`. `path` gets the point across, so this is optional.

Shared/DsaUtility.cpp
```diff
@@ -41,23 +50,95 @@ namespace Dsa {
...
+      {
+        if (configurationNode.isObject())
+        {
+          // get the selected value
+          const auto configurationObject = configurationNode.toObject();
+          const auto configurationSelected = configurationObject["selected"].toBool();
+          if (configurationSelected)
+          {
+            selectedConfigurationName = configurationObject["name"].toString();
+            break;
```
JamesMBallard, line 122: Some heavy nesting in this method as well. It may be worse to try to refactor it (it looks necessary). See if can be simplified without making readability suffer.

JamesMBallard (commented):
Thanks for your patience with the review. I did not review anything in the 3rdparty folder.

Shared/configurations/ConfigurationListModel.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+  beginInsertRows(QModelIndex{}, m_configurations.count(), m_configurations.count());
+  m_configurations.emplaceBack(name, url, selected, loaded, percentDownloaded);
+  endInsertRows();
+  return true;
+}
+
+QHash<int, QByteArray> ConfigurationListModel::roleNames() const
+{
+  static QHash<int, QByteArray> roles = {
```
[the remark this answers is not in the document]
dtedder: added const

Shared/configurations/ConfigurationListModel.cpp
```diff
@@ -0,0 +1,142 @@
...
+      configuration.setUrl(value.toString());
+      break;
+    case Roles::Selected:
+      configuration.setSelected(value.toBool());
+      break;
+    case Roles::PercentDownloaded:
+      configuration.setPercentDownloaded(value.toInt());
+      break;
+    default:
+      break;
```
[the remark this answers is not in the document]
dtedder: removed

Shared/configurations/ConfigurationListModel.cpp
```diff
@@ -0,0 +1,142 @@
...
+    case Roles::Downloaded:
+      return configuration.downloaded();
+    case Roles::Selected:
+      return configuration.selected();
+    case Roles::Loaded:
+      return configuration.loaded();
+    case Roles::PercentDownloaded:
+      return configuration.percentDownloaded();
+    default:
+      return QVariant{};
```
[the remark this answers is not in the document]
dtedder: removed

Shared/configurations/ConfigurationController.h
dtedder: switched

Shared/configurations/ConfigurationListModel.cpp
dtedder: removed per our discussion in teams
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#c30-define-a-destructor-if-a-class-needs-an-explicit-action-at-object-destruction

Shared/configurations/ConfigurationController.cpp
dtedder: indeed there were many places here that should have been early returns or continues. it has been reduced to only 3 indents.

Shared/configurations/ConfigurationController.cpp
dtedder: added blank line and made comments in this section a bit more clear

Shared/configurations/ConfigurationController.cpp
dtedder: Not needed. Confirmed in the Qt source, `close()` is called in the parent class QFileDevice destructor, which itself calls `flush()`.

Removed.

Shared/configurations/ConfigurationController.cpp
dtedder: Agreed, this should be avoided. I've added extra line for clarity and updated comments where appropriate.

Shared/configurations/ConfigurationController.cpp
dtedder: 👍 Added guard

Shared/configurations/ConfigurationController.cpp
dtedder: This method comes from the AbstractTool class. For some reason, was thinking it was pure virtual and needed to implemented in the new controller I created for configuration. This controller actually manages a resource that is 'above' the AppConfig and thus doesn't need to store or read anything from the properties object that's passed to/from. I was ignoring before but will remove since there's no need for it.

Shared/configurations/ConfigurationController.cpp
dtedder: The current implementation does not yet support managing multiple configurations with the application itself. The configurations json file and the data it points to must be managed manually (aside from the Default configuration from our AGO item). We hope to soon add new abilities for users to add/download/delete their own. TODO

Shared/configurations/ConfigurationController.cpp
dtedder: 👍

Shared/configurations/ConfigurationController.cpp
dtedder: Added this in the constructor for the QNetworkManager. I am using a pointer for the 'active' networkReply. Should I continue to set = nullptr ? This pointer is checked in other places to guard against null but perhaps this makes that not necessary either?

Shared/configurations/ConfigurationController.cpp
dtedder: 👍

Shared/configurations/Configuration.h
dtedder: removed

Shared/configurations/Configuration.h
dtedder: added

Shared/configurations/ConfigurationController.cpp
JamesMBallard: I didn't realize there was a constructor that allows setting this value.

>Should I continue to set = nullptr ? This pointer is checked in other places to guard against null but perhaps this makes that not necessary either?

Hmm, does it need to be a member? In our codebase we don't store any QNetworkReply objects since they're ephemeral and there are other ways to keep track of them.

Generally it's enough that we capture the reply in any lambda connected to the `finished` signal, which we know is the end of the line - nothing happens after that point, and the QNetworkAccessManager is free to delete it. Is it possible to refactor like that? If not, you are right you're going to have some lifetime issues to solve, or you could stick with what you have now. Your code isn't wrong or bad, and it does seem that this one improvement could necessitate more changes elsewhere. I am ok either way.

Shared/DsaUtility.cpp
dtedder: There isn't a situation at the moment where these values should be changed that I am aware of. I think if they did, it would cause a crash in the app. I added this logic because the introduction of the Configurations feature lead to this part of the DsaUtilities app needing to be reorganized a bit. And ultimately these methods will now be called multiple times. And because they shouldn't change, it made sense to cache the resolved paths.

I will give them better names.

Shared/DsaUtility.cpp
dtedder: Applied same refactoring from above. Early returns and continues reduced nesting to a reasonable level.

GuillaumeBelz:
For 3rdparty libs, I think it's better to move that in sub directory, to be more explicit. `3rdparty/zlib-ng`. 

I didn't check which part(s) of zlib is used in this PR, but it's probably better to provide only used files.

Handheld/Handheld.pro
```diff
@@ -24,13 +24,16 @@ ARCGIS_RUNTIME_VERSION = 200.5.0
 DEFINES += ARCGIS_MAPS_SDK_VERSION=$$ARCGIS_RUNTIME_VERSION
 include($$PWD/../Shared/build/arcgisruntime.pri)
 
+include($$PWD/../3rdparty/zlib_minizip_ng.pri)
```
GuillaumeBelz, line 27: Except if I miss something, the path is not correct. The zlib files are in `3rdparty`

dtedder:
@GuillaumeBelz moved zlib files into their own folder
👍
https://github.com/Esri/dynamic-situational-awareness-qt/pull/399/commits/efb6a21fdf0b1991c9715c8570dee2159c6228cd

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,415 @@
...
+    auto configurationDirectoryPath = dirConfigurations.filePath(configurationName);
+    QDir dirConfiguration{configurationDirectoryPath};
+
+    // skip directories that are not found on disk
+    if (!dirConfiguration.exists())
+      continue;
+
+    // skip if the app config cannot be found in the directory
+    QFile fileSettings{dirConfiguration.filePath("DsaAppConfig.json")};
+    if (!fileSettings.exists())
```
JamesMBallard, line 380: Prefer the static QFile::exists for something like this. No need to create an object and have it hanging around.
https://doc.qt.io/qt-6/qfile.html#exists

Shared/qml/Options.qml
```diff
@@ -19,224 +19,398 @@ import QtQuick.Controls
...
+                        //     width: 0 // TODO: remove when downloads from user are implemented
+                        //     anchors {
+                        //         right: parent.right
+                        //         verticalCenter: parent.verticalCenter
+                        //     }
+                        //     MouseArea {
+                        //         anchors.fill: parent
+                        //         onClicked: configurationController.remove(index)
+                        //     }
+                        // }
```
JamesMBallard, line 376: Can this be removed?

JamesMBallard (commented):
Looking pretty good overall. A few comments.

Handheld/qml/main.qml
```diff
@@ -599,4 +613,16 @@ Handheld {
...
+    DsaYesNoDialog {
+        id: configurationDownloadDialog
+        informativeText: "Download the default configuration data from Esri (~450mb)?"
+        onAccepted: showConfigurations(true);
+        onRejected: showConfigurations(false);
+    }
+    function showConfigurations(downloadDefaultData) {
+        if (downloadDefaultData)
+            configurationController.downloadDefaultData();
+        optionsTool.showConfigurationsTab();
```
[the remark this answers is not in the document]
dtedder: Fixed both of these.
ec02535080f56ce1690e7165eec2f83a3dad3eea

Shared/DsaUtility.h
```diff
@@ -33,7 +33,12 @@ class DsaUtility
 {
 public:
 
-  static QString dataPath();
+  inline static const QString FILE_NAME_DSA_CONFIGURATIONS{"DsaConfigurations.json"};
+  inline static const QString FILE_PATH_HOME_DATA{"ArcGIS/Runtime/Data/DSA"};
```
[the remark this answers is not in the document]
dtedder: Can't find a way to make constexpr will work here with QString because it is a non-literal type. Could switch to this:
`static constexpr char FILE_NAME_DSA_CONFIGURATIONS[] = "DsaConfigurations.json";`
But that doesn't seem as clean.

I would probably be ok with this:
`static constexpr char* FILE_NAME_DSA_CONFIGURATIONS = "DsaConfigurations.json";`
But this raises the warning as of C++ 11 about converting from string literal to const char*.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,415 @@
...
+{
+  m_configurationListModel->select(index);
+  storeConfigurations();
+  emit configurationsChanged();
+}
+
+void ConfigurationController::download(int index)
+{
+  // prevent calling the download function if something is already in progress
+  if (m_downloading)
```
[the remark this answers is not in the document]
dtedder: I like `downloadInProgress` better.

7ea65c3629335d5522369e2aa7fee28b5701b184

Shared/qml/Options.qml
dtedder: Planning to enable this when time permits and users have the capability to add/download their own data to devices. I added a TODO but it's a bit hidden. I moved to the top to make more visible.
e8f91d09c4f87e6c151957afbcae0bd989af6067

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,415 @@
...
+  QJsonArray array;
+  for (const auto& cfg : *m_configurationListModel)
+  {
+    QJsonObject obj;
+    obj["selected"] = cfg.selected();
+    obj["name"] = cfg.name();
+    obj["url"] = cfg.url();
+    array.append(obj);
+  }
+  QJsonDocument doc{array};
```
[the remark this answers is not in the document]
dtedder: added const
e8f91d09c4f87e6c151957afbcae0bd989af6067

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,415 @@
...
+  if (error != QNetworkReply::NetworkError::OperationCanceledError)
+    emit toolErrorOccurred(QString("Network Error (%1)").arg(error), "Error Downloading");
+}
+
+void ConfigurationController::extractCompleted()
+{
+  // delete the temporary download from the device
+  QFile::remove(m_downloadFileName);
+  setPercentComplete(100);
+  m_zipHelper->deleteLater();
```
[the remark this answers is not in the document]
dtedder: With the current approach, I don't think unique_ptr will eliminate the need to clean this resource up explicitly. The class design right now is just using a single 'download'(QNetworkManager) and 'unzip'(ZipHelper). When multiple downloads are possible with the app, it will make more sense to create the object using std::make_unique and add it to a vector or other container that can manage it's lifetime. In the completed lambda, the element can be popped off the container which will clean it up. But I think perhaps this should be added after the deleteLater calls:
`m_zipHelper = nullptr;`?

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,415 @@
...
+    if (!fileSettings.exists())
+      continue;
+
+    // check the folder recursively for other files
+    QDirIterator dirIt(configurationDirectoryPath, QDir::Files | QDir::NoDotAndDotDot, QDirIterator::Subdirectories);
+    auto filesInConfigurationFolder = 0;
+    while (dirIt.hasNext())
+    {
+      filesInConfigurationFolder++;
+      QFile f{dirIt.next()};
```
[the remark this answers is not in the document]
dtedder: It is not used. This was just called here to advance the iterator to the next file to continue checking for files. I overlooked .next(). It returns an empty QString at the end. So that can be used to check and advance in the while clause since we don't really care about the returned file.

Better approach added here:
fc45dc60db82a7ab5d036457fec62f9ecb4b4697

Shared/configurations/ConfigurationController.cpp
dtedder: fc45dc60db82a7ab5d036457fec62f9ecb4b4697

Shared/configurations/ConfigurationController.cpp
```diff
@@ -0,0 +1,415 @@
...
+  if (!dirActiveDownloadConfiguration.exists())
+    dirConfigurations.mkdir(activeDownloadName);
+
+  // extract the downloaded file to the configuration folder
+  m_zipHelper->extractAll(dirActiveDownloadConfiguration.absolutePath());
+}
+
+void ConfigurationController::downloadErrorOccurred(QNetworkReply::NetworkError error)
+{
+  m_aborted = true;
```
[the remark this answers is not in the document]
dtedder: There is distinct behavior for user-cancelled vs network error. Network error will display a message to the user while user-cancelled will not.

The aborted flag however is used for both scenarios in order to attempt to clean up any remaining download file as well as setting the percent downloaded properties of the model for the download to 0.

My logic here was that in both cases, the operation was aborted. How it was aborted is a separate detail and is taken care of by showing or not showing the alert message.

Shared/configurations/ConfigurationController.cpp
dtedder: At the moment, no I don't think I will be able to make use of the autoDeleteReplies feature of QNetworkManager. Hopefully in the short term, I will be able to switch to the multiple download/extract strategy. At that point, the Dsa::Configuration instances can be set to be the parent of the QNetworkReply and autoDeleteReplies can be used.

Shared/configurations/ConfigurationController.cpp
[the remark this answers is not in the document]
JamesMBallard: >I don't think unique_ptr will eliminate the need to clean this resource up explicitly

No, it won't. You are right. Yes, `m_zipHelper = nullptr;` should be added to anywhere you're calling `deleteLater`. Unique_ptr would allow you to simply call `reset()` to delete it and avoid the `= nullptr` part, but it's not worth changing it for that alone.

Shared/configurations/ConfigurationController.cpp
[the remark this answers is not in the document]
JamesMBallard: Ok, that makes sense.

Shared/configurations/ConfigurationController.cpp
dtedder: updated here:
c8dbcf7ccaa8878d09aa3d6d5a137ec98e7e2945

Shared/DsaUtility.h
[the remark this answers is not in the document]
dtedder: Yep. Updated here:
44b7fc2c1fb247d9e8ca23233386362426301de3

Shared/configurations/ConfigurationController.cpp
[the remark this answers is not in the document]
dtedder: added these = nullptr lines for the dangling pointer.
1e64798cd9ed6459203840fba3091a3463dba5ef

Issue #410, opened 2025-02-06
Create sanity test for DSA


kerryrobinson:
Create a manual test plan for DSA and add to repo. 

Leverage tests captured here: https://devtopia.esri.com/runtime/qt-test/issues?q=is%3Aissue+label%3AW-DefenseAndIntel+

PR #419, opened 2025-02-18
410 create sanity test for dsa

kerryrobinson:
#410 

Sanity test markdown created and formatted.  There might be a few minor things still to update (including a section on the new configuration download tool), but this is good enough for now.  We'll have to remember to add a section when we add significant functionality.


dtedder (approved):
👍

PR #420, opened 2025-02-18
Dtedder/migration doc updates2

dtedder:


README.md
```diff
@@ -103,10 +104,6 @@ To quickly obtain these additional `*.stylx` files, we provide these in the [dem
 
 Esri welcomes contributions from anyone and everyone. Please see our [guidelines for contributing](https://github.com/esri/contributing).
 
-## MDTOC
-
-Generating table of contents for documents in this repository was performed using the [MDTOC package for Atom](https://atom.io/packages/atom-mdtoc).
-
 ## Licensing
 
 Copyright 2018 Esri
```
kerryrobinson, line 109: Should we update this to 2025?

docs/DataManagement.md
```diff
@@ -0,0 +1,33 @@
+- Migrating DSA data used prior to the 2.0.0 release
+
+The DSA application has a base file structure at which the mapping and configuration data is expected to exist. Prior to the 2.0.0 release of DSA, this location was <%home%>/ArcGIS/Runtime/Data/DSA. At the 2.0.0 release, a new feature was introduced in order to give users the ability to manage multiple configurations for the application. For example, you may wish to use the Esri provided dataset. But you may also have data for a relevant geographic location that covers your area of operations. Prior to the introduction of 'Configurations' the user would need to shutdown the application and copy, move or strategically rename folders in order to load a new set of data and application settings. The Configuration feature adds a new folder level at the very bottom of the structure to allow users to host more than 1 set of DSA data at a time.
+
+The following dialog describes the migration process for MacOS. But the same steps should be followed for Windows, Linux or Android.
```
kerryrobinson, line 5: I think we should mention here that to make updates to file storage on Android, because it no longer uses shared storage, you must use Android Studio to manage data. Actually, Android users prior to 2.0.0 will need to move it from the older shared location to the newer location.   So, adding a sentence here that specifically mentions that will be helpful.

kerryrobinson (changes requested):
All looks good and the migration text reads well. This'll be super helpful. Just suggested a few comments and some formatting changes. You can adjust the width of the app screenshots to whatever size looks best.

README.md
dtedder: As I have been working with James on the repo so far, we only typically update the Copyright date on new files as they are added. After a bit of research on licensing/copyright statements in open source projects, it seems there's a consensus that having the original in place is typically good enough. And some argue it's not technically necessary at all with the detailed audit trail that is built into platforms like github already. I think we can leave it as is for now.

docs/DataManagement.md
dtedder: Added a message for this for Android.

README.md
```diff
@@ -30,6 +27,10 @@ The DSA apps do not require access to a server. All the data and processing are
 
 Visit the DSA [Open Source App page] for more details.
```
kerryrobinson, line 28:
```suggestion

```
We should remove this sentence since the link no longer works.

kerryrobinson (approved):
OK, just one more line to remove in the readme.   Otherwise all good! 👍

PR #424, opened 2025-04-04
Check app executable location for Configuration file

dtedder:
adding a check for the DsaConfigurations.json file within the same folder as the application executable on desktop (Linux/MacOS/Windows) platforms.

### DIFF
--- a/Shared/DsaUtility.cpp
+++ b/Shared/DsaUtility.cpp
@@ -31,6 +31,7 @@
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QStandardPaths>
+#include <QtGlobal>
 #include <QtSystemDetection>
 #include <QJsonValueConstRef>
 #include <QtMath>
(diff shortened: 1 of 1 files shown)


Shared/DsaUtility.cpp
```diff
@@ -55,6 +58,17 @@ QString DsaUtility::configurationsDirectoryPath()
   if (!fullPathToConfigurationsDirectory.isNull())
     return fullPathToConfigurationsDirectory;
 
+#if defined(Q_OS_LINUX) || defined(Q_OS_MACOS) || defined(Q_OS_WINDOWS)
+  // if running on a desktop system check the current application directory for the configurations file
+  QDir appDir{QCoreApplication::applicationDirPath()};
+  QFile configFile{appDir.absoluteFilePath(FILE_NAME_DSA_CONFIGURATIONS)};
+  if (configFile.exists())
```
JamesMBallard, line 65: Can you use `QFile::exists()` (static version) ?

Shared/DsaUtility.cpp
```diff
@@ -55,6 +58,17 @@ QString DsaUtility::configurationsDirectoryPath()
   if (!fullPathToConfigurationsDirectory.isNull())
     return fullPathToConfigurationsDirectory;
 
+#if defined(Q_OS_LINUX) || defined(Q_OS_MACOS) || defined(Q_OS_WINDOWS)
+  // if running on a desktop system check the current application directory for the configurations file
+  QDir appDir{QCoreApplication::applicationDirPath()};
```
JamesMBallard, line 63:
```suggestion
  const QDir appDir{QCoreApplication::applicationDirPath()};
```

Shared/DsaUtility.cpp
```diff
@@ -55,6 +58,17 @@ QString DsaUtility::configurationsDirectoryPath()
   if (!fullPathToConfigurationsDirectory.isNull())
     return fullPathToConfigurationsDirectory;
 
+#if defined(Q_OS_LINUX) || defined(Q_OS_MACOS) || defined(Q_OS_WINDOWS)
```
JamesMBallard, line 61: I would consider restructuring some of this logic. Down below there is already `#ifdef Q_OS_IOS`, `#else`, then `#endif`.

Android logic is there but not explicit.

Would you consider making the branching more explicit?

For example:
```c++
#if defined(Q_OS_LINUX) || defined(Q_OS_MACOS) || defined(Q_OS_WINDOWS)
  // ...
#elif defined(Q_OS_IOS)
  // ...
#elif defined(Q_OS_ANDROID)
  // ...
#endif 
```

Shared/DsaUtility.cpp
```diff
@@ -23,12 +23,15 @@
...
+#include <QCoreApplication>
 #include <QDir>
+#include <QFile>
 #include <QFileInfo>
 #include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QStandardPaths>
+#include <QtSystemDetection>
 #include <QJsonValueConstRef>
```
JamesMBallard, line 35: This is probably overkill, but QtGlobal should be included any time the Qt platform ifdefs are used. Please toss it in here. I am aware this was already an oversight here, but might as well clean it up.

dtedder: 👍 added

Shared/DsaUtility.cpp
dtedder: 👍 Instance of QFile was not really necessary here. Switched.

Shared/DsaUtility.cpp
dtedder: Added const

Shared/DsaUtility.cpp
dtedder: rearranged this logic to be more explicit.

If iOS, Documents directory is required.
Else if Android, Home directory is required.
Else check the executable's location for the DsaConfigurations.json file. If it exists there, the app directory can immediately be returned as the starting directory. Otherwise, Home directory is required.

dtedder: The set of checks for `defined(Q_OS_LINUX) || defined(Q_OS_MACOS) || defined(Q_OS_WINDOWS)` actually ended up not working very well because Android and Linux overlap.

JamesMBallard: Good catch, I forgot about that.

PR #432, opened 2025-05-02
Add ZOffset and SurfacePlacement to config file

### LINKED ISSUE #365: Cursor Offset
Currently, the cursor offset is 10m above ground surface and in some scenarios maybe too much.
It would be good to make an option in the config file to define what the surface placement should be (draped, relative, fixed) plus an offset value. To have this set up handy in the DSA app settings, would be ideal.

dtedder:
Configuration parameters have been added to the default configuration for a Z offset value and the scene SurfacePlacement property for the Location3d layer. Users can specify 'DrapedFlat' to always have the marker symbol drawn onto the terrain, Z offset will be ignored. Any other value will result in SurfacePlacement::Relative and the Z offset will be used for the offset value.

#365

### DIFF
--- a/3rdparty/zlib-ng/zutil.h
+++ b/3rdparty/zlib-ng/zutil.h
@@ -103,7 +103,7 @@ extern z_const char * const PREFIX(z_errmsg)[10]; /* indexed by 2-zlib_error */
 #  define OS_CODE  6
 #endif
 
-#if defined(MACOS) || defined(TARGET_OS_MAC)
+#if defined(MACOS)
 #  define OS_CODE  7
 #endif
 
--- a/Shared/LocationController.cpp
+++ b/Shared/LocationController.cpp
@@ -35,6 +35,7 @@
 #include "ModelSceneSymbol.h"
 #include "Point.h"
 #include "SceneQuickView.h"
+#include "SceneViewTypes.h"
 #include "SpatialReference.h"
 #include "SymbolTypes.h"
 
@@ -49,10 +50,6 @@
 using namespace Esri::ArcGISRuntime;
 
 namespace Dsa {
-
-const QString LocationController::SIMULATE_LOCATION_PROPERTYNAME = "SimulateLocation";
-const QString LocationController::GPX_FILE_PROPERTYNAME = "GpxFile";
-const QString LocationController::RESOURCE_DIRECTORY_PROPERTYNAME = "ResourceDirectory";
 
 /*!
   \class Dsa::LocationController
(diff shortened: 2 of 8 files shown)


3rdparty/zlib-ng/zutil.h
```diff
@@ -103,7 +103,7 @@ extern z_const char * const PREFIX(z_errmsg)[10]; /* indexed by 2-zlib_error */
 #  define OS_CODE  6
 #endif
 
-#if defined(MACOS) || defined(TARGET_OS_MAC)
+#if defined(MACOS)
```
JamesMBallard, line 106: Was this intentional?

Shared/LocationController.cpp
```diff
@@ -205,10 +202,29 @@ QString LocationController::toolName() const
...
+  // skip if the location display was not ready
+  if (!m_locationDisplay3d)
+    return;
+
+  // make sure the value from the configuration file for
+  // the z offset is able to be converted to a double
+  bool ok;
+  double offset = properties[PROPERTY_NAME_CURRENT_LOCATION_Z_OFFSET].toString().toDouble(&ok);
+  if (ok)
+    m_locationDisplay3d->setZOffset(offset);
```
JamesMBallard, line 220: There may be a more direct way to check this. How about something like this?

```c++
if (const auto zOffsetVar = properties[PROPERTY_NAME_CURRENT_LOCATION_Z_OFFSET]; zOffsetVar.canConvert<double>())
{
  m_locationDisplay3d->setZOffset(zOffsetVar.toDouble());
}
```

Your way should work as well but this would avoid converting to a string first and then again to a double.

3rdparty/zlib-ng/zutil.h
dtedder: This was intentional, yes. I started getting the warning for redefined macro I believe as a result of updating Xcode on my machine to 16 but it's hard to know for sure. I know this wasn't happening before the 2.0.0 release.

I was able to track down the same commit in the zlib-ng repository. https://github.com/zlib-ng/zlib-ng/commit/153b9648ed1bfde4c2a8f76844a50620c3b21b21.

But the warning shows up because there is also a match below with `__APPLE__` below and the OS_CODE variable is defined again.

JamesMBallard: Ok, it sounds fine but our typical policy is to leave 3rd party code unchanged. I would rather just disable warnings from this code if possible. The more we tweak the harder it becomes to upgrade it all later on without risk of losing all our little improvements like this.

Shared/LocationController.cpp
dtedder: Makes sense. Pushed up a commit to reduce the conversions.

Shared/LocationController.cpp
```diff
@@ -205,10 +202,29 @@ QString LocationController::toolName() const
...
+
+  // allow for the option to use draped-flat style surface placement, otherwise default to relative
+  auto surfacePlacementValue = properties[PROPERTY_NAME_CURRENT_LOCATION_SURFACE_PLACEMENT].toString();
+  if (surfacePlacementValue == QStringLiteral("DrapedFlat"))
+    m_locationDisplay3d->setSurfacePlacement(SurfacePlacement::DrapedFlat);
+  else
+    m_locationDisplay3d->setSurfacePlacement(SurfacePlacement::Relative);
```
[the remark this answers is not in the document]
dtedder: 👍 . Pushed commit for it.

JamesMBallard (approved):

Issue #431, opened 2025-05-01
Remove armeabi-v7a and x86 architecture from Android.pri file


har13205:
While building the app on Android on a `arm64-v8a ` device, we are facing the following error as qmake no longer supports multi-arch builds 

`Cannot find application binary in build dir C:/dynamic-situational-awareness-qt/Handheld/build/Android_Qt_6_8_2_Clang_arm64_v8a-Debug/android-build//libs/armeabi-v7a/libDSA_Handheld_Qt_armeabi-v7a.so `

To avoid this, we need to remove the armeabi-v7a  and x86 from the [Android.pri](https://github.com/Esri/dynamic-situational-awareness-qt/blob/050513e983a9d66098132f2faf0eb6dd231fed0f/Handheld/Android/Android.pri#L32) file

dtedder:
@kerryrobinson 

This can be checked by doing a build for Android (vehicle/handheld) and ensuring you have checked out to `v.next` locally with no modifications to any tracked files. You should not get the compilation error we have seen for missing libraries for the arm v7 and x86 architectures.

You should also no longer see any warnings related to gradle. These are harder to spot but end up in the Compile Output window of Qt Creator.

kerryrobinson:
Confirmed working as expected.

PR #435, opened 2025-05-02
Dtedder/minor cleanup

dtedder:
Did some cleanup of the Android specific project files. Including removing architectures other than `arm64-v8a`.

Also fixed the typo in the default constructor for `Configuration` giving the warning.

#431

### DIFF
--- a/Shared/configurations/Configuration.h
+++ b/Shared/configurations/Configuration.h
@@ -30,11 +30,6 @@ public:
                 bool selected,
                 bool loaded,
                 int percentDownloaded);
-  Configuration() = default;
-  Configuration(const Configuration&) = default;
-  Configuration(Configuration&&) = default;
-  Configuration& operator=(const Configuration&) = default;
-  Configuration& operator=(Configuration&&) = default;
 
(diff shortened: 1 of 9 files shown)


Shared/configurations/Configuration.h
```diff
@@ -30,7 +30,7 @@ class Configuration
                 bool selected,
                 bool loaded,
                 int percentDownloaded);
-  Configuration(Configuration&) = default;
+  Configuration() = default;
```
JamesMBallard: Was this intended? I wouldn't think a Configuration with all default values is useful.

```suggestion
  Configuration() = delete;
```

dtedder: I did some more testing and realized there's no need for any of these special method definitions. I only added these originally because `Configuration` is used in a container. I was certain I was getting a warning or error related to the call to `emplace_back` in `ConfigurationListModel`... No longer able to reproduce it though. Maybe simpler to just remove them and leave it to the compiler.

JamesMBallard (approved):

PR #437, opened 2025-05-06
Surface pointer not updated on Elevation Source Selected signal

dtedder:
The member variable for the surface was not getting updated when the local data was added as an elevation source which caused an exception to be thrown which was not being caught. This connects the LocationTextController to the AddLocalDataController's signal so it gets updated properly.

It also adds an exception continuation to the elevationAsync in the case where the app is misconfigured or an invalid elevation source is added. After 3 attempts, the pointer will be set to null and will stop attempting to get an elevation until a new elevation source is selected.

### DIFF
--- a/Shared/LocationTextController.cpp
+++ b/Shared/LocationTextController.cpp
@@ -159,7 +159,6 @@ void LocationTextController::onLocationChanged(const Point& pt)
 
 void LocationTextController::onToolAdded(AbstractTool* newTool)
 {
-  // skip if the tool pointed to is not valid
   if (!newTool)
     return;
 
@@ -322,7 +321,6 @@ void LocationTextController::formatElevationText(double elevation)
 
 void LocationTextController::resetSurface()
 {
-  // update the current surface pointer
   if (const auto* scene = ToolResourceProvider::instance()->scene(); scene)
   {
     m_elevationFailures = 0;
--- a/Shared/LocationTextController.cpp
+++ b/Shared/LocationTextController.cpp
@@ -21,6 +21,7 @@
 
 // C++ API headers
 #include "CoordinateFormatter.h"
+#include "ErrorException.h"
 #include "Scene.h"
 #include "Surface.h"
 
@@ -28,6 +29,7 @@
 #include <QFuture>
 
 // DSA headers
+#include "AddLocalDataController.h"
 #include "ToolManager.h"
 #include "ToolResourceProvider.h"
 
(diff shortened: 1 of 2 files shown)


Shared/LocationTextController.cpp
```diff
@@ -128,25 +132,49 @@ void LocationTextController::onLocationChanged(const Point& pt)
...
-    auto* surface = scene->baseSurface();
-    if (!surface)
-      return;
-
-    // set the elevation text in the continuation block
-    surface->elevationAsync(pt).then(this, [this](double elevation)
+    Q_UNUSED(e);
+    // multiple failures likely indicate a failed loading of the
+    // elevation source for the surface. set the member to nullptr
+    // to prevent continued calls for fetching the elevation
```
JamesMBallard, line 153: Would it be preferable to just wait to call `m_surface->elevationAsync` once the ElevationSource is loaded?

Try three times seems brittle.

Shared/LocationTextController.cpp
```diff
@@ -128,25 +132,49 @@ void LocationTextController::onLocationChanged(const Point& pt)
...
+    m_elevationFailures++;
+    if (m_elevationFailures > 2)
+      m_surface = nullptr;
```
JamesMBallard, line 156: Is there any point in resetting `m_elevationFailures` here as well if we get into the condition? Assuming we keep this logic.

Shared/LocationTextController.cpp
```diff
@@ -128,25 +132,49 @@ void LocationTextController::onLocationChanged(const Point& pt)
...
+  });
+}
+
+void LocationTextController::onToolAdded(AbstractTool* newTool)
+{
+  // skip if the tool pointed to is not valid
```
JamesMBallard:
```suggestion
```

I certainly hope that is obvious :-)

Shared/LocationTextController.cpp
```diff
@@ -296,6 +320,16 @@ void LocationTextController::formatElevationText(double elevation)
   emit currentElevationTextChanged();
 }
 
+void LocationTextController::resetSurface()
+{
+  // update the current surface pointer
```
JamesMBallard:
```suggestion
```

This is subjective, but I do think most developers can tell what's happening here.

Shared/LocationTextController.cpp
dtedder: OCD. Blocks of code with no comment just seem to bug me, even tiny self explanatory ones like this. Usually I go back and drop them in SourceTree before I commit.

Shared/LocationTextController.cpp
dtedder: Removed.

Shared/LocationTextController.cpp
dtedder: Yes, this doesn't seem right after looking into it more. I'll submit a new PR with a better approach based on loading the ElevationSource first.

PR #438, opened 2025-05-07
Dtedder/390 fix elevation source loading

### LINKED ISSUE #390: Update member pointer to Surface on LocationTextController on change
When a user adds a new data source and uses it for elevation, there is no code to set the m_surface pointer on the LocationTextController tool. The current code only updates this property on GeoView changed which is not triggered by data getting added in this way. As a work around, the location changed event is currently querying to find the baseSurface every time. This should be updated by the SceneQuckView::layerViewStateChanged signal or by a custom signal that can be raised by the AddLocalDataController tool.

dtedder:
fix issue with surface pointer not getting updated if the app is started with an elevation source in the config file that is not valid.

### DIFF
--- a/Shared/AddLocalDataController.cpp
+++ b/Shared/AddLocalDataController.cpp
@@ -238,31 +238,38 @@ void AddLocalDataController::addItemAsElevationSource(const QList<int>& indices)
 */
 void AddLocalDataController::createElevationSourceFromTpk(const QString& path)
 {
-  TileCache* tileCache = new TileCache(path, this);
+  const auto* scene = ToolResourceProvider::instance()->scene();
+  if (!scene)
+    return;
 
-  connect(tileCache, &TileCache::doneLoading, this, [this, tileCache](Error error)
+  auto* cache = new TileCache(path, this);
+  connect(cache, &TileCache::errorOccurred, this, &AddLocalDataController::errorOccurred);
+  connect(cache, &TileCache::doneLoading, this, [this, scene, cache](const Error& errorCache)
   {
-    if (!error.isEmpty())
+    if (!errorCache.isEmpty())
       return;
 
-    // Check if the tiles are LERC encoded
-    if (tileCache->tileInfo().format() == TileImageFormat::LERC)
+    // tiles must be LERC encoded
+    if (cache->tileInfo().format() != TileImageFormat::LERC)
+      return;
+
+    // create new elevation source from the tile cache
+    auto* source = new ArcGISTiledElevationSource(cache, this);
+    connect(source, &ArcGISTiledElevationSource::errorOccurred, this, &AddLocalDataController::errorOccurred);
+    connect(source, &ArcGISTiledElevationSource::doneLoading, this, [this, cache, scene, source](const Error& errorSource)
     {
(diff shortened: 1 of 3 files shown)


Shared/AddLocalDataController.cpp
```diff
@@ -238,47 +238,62 @@ void AddLocalDataController::addItemAsElevationSource(const QList<int>& indices)
...
+    if (cache->tileInfo().format() != TileImageFormat::LERC)
+      return;
 
-      auto scene = ToolResourceProvider::instance()->scene();
-      if (scene)
-        scene->baseSurface()->elevationSources()->append(source);
+    // create new elevation source from the tile cache
+    auto* source = new ArcGISTiledElevationSource(cache, this);
+    connect(source, &ArcGISTiledElevationSource::errorOccurred, this, &AddLocalDataController::errorOccurred);
+    connect(source, &ArcGISTiledElevationSource::doneLoading, this, [this, cache, scene, source](const Error& errorSource)
```
JamesMBallard, line 259: This is optional, but for connections for things that only happen once (like doneLoading), you can consider a [SingleShotConnection](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum). That will release any resources held by the lambda when it's no longer needed.

I don't think it will cause any issues, but unclear if it's worth changing. Fair warning it's not a pattern we use very often, but I can see it being helpful.

Shared/AddLocalDataController.cpp
```diff
@@ -238,47 +238,62 @@ void AddLocalDataController::addItemAsElevationSource(const QList<int>& indices)
...
+        return;
 
+      scene->baseSurface()->elevationSources()->append(source);
       emit elevationSourceSelected(source);
-      emit propertyChanged(DEFAULT_ELEVATION_PROPERTYNAME, tileCache->path());
-    }
+      emit propertyChanged(DEFAULT_ELEVATION_PROPERTYNAME, cache->path());
+    });
+
+    source->load();
```
JamesMBallard, line 269: To be sure I understand the nature of the fix, it is that now we explicitly load the `ArcGISTiledElevationSource`, and only emit `propertyChanged(DEFAULT_ELEVATION_PROPERTYNAME...` once it has completed loading. Does that sound correct?

That way downstream logic can remain unchanged and doesn't need to worry about whether it is loaded or not.

dtedder: Correct. Now the downstream logic in the location changed update can stay what it was before. Just check to make sure `m_surface` pointer is not null. The logic added here will prevent invalid elevation sources from ever being added (which is what was throwing the exceptions).

Shared/AddLocalDataController.cpp
dtedder: I'll take a look at this. Sounds useful for cases like these.

JamesMBallard: On second thought I think we should not use it. There could be too many gotchas where we might guess wrong about a signal that only emits once. It's harder to reason about than I realized.

JamesMBallard (approved):

dtedder:
previous PR (#437), closed in favor of this approach

Issue #418, opened 2025-02-18
Packages tool (Handheld): Overlays list does not show list of layers from the MSPK


kerryrobinson:
When opening an MSPK using the Handheld version of the app on Android, the MSPK opens correctly, but the Overlays tool still shows the original layers that were added to the scene prior to opening the MSPK.  Those layers are not being drawn the scene, just listed in the Overlays panel. 

Correct behavior is that it shows the list of layers from the MSPK in the Overlays list (the original layers should have been removed from the scene). 

https://github.com/user-attachments/assets/36d8ace8-b8bc-4f9c-8b68-cce9bd656b0c

dtedder:
@har13205 branch is ready for builds

@kerryrobinson some things to check:
- [x] starting with clean configuration, user should be able to select a package from the device, close the app, reopen without a crash
- [ ] Overlays and Message Feeds panels should update accordingly with the active package or the default scene no matter how many packages are opened, reopened

har13205:
@kerryrobinson builds are placed at `\\apps-data\Data\apps\DSA\builds\Issue-418`

kerryrobinson:
@dtedder I am still able to reproduce this and noticed that is particular to running on Android. I reproduced it with both the Handheld and Vehicle app on my Android device.  Essentially, once you load an MSPK, it does not display the layers from the MSPK in the Overlays panel. If there are already layers in the Overlays panel before loading the MSPK, they remain in the Overlays panel but no in the Scene. Turning those layers on/off has no effect, and trying to remove them from the list has no effect. 

I do not experience this on Windows.  I also am able to close the app after loading an SLPK and repoen it without crash.

dtedder:
@kerryrobinson confirmed this also happens with Handheld on iOS. I created a one off build pointing to a test item that included a MSPK in the 'default' download. It does show that the layer list is not updated. Closing and re-opening works on iOS as well and the layers will be correct on the subsequent restart of the app.

kerryrobinson:
@dtedder  I tested running the Vehicle app on Windows with the smaller form factor and was unable to reproduce the problem. 

I tested an older version of the Handheld app (1.6) and was able to reproduce the problem on Android.

dtedder:
Unblocking this issue. There are potential changes coming in the development work related to #456 that may resolve this issue. The UI needs to be refreshed whenever the list has been modified. Will begin resting this issue on that branch.

har13205:
[...]

PR #442, opened 2025-05-15
CurrentPackage causes issues on startup

### LINKED ISSUE #450: CurrentPackage causes crash on app restart
Modifications to the application startup procedure related to the conversion from GraphicOverlay to DynamicEntity based Message Feeds caused a crash upon restart of the application if a MobileScenePackage is selected. A race condition existed in the setupFeeds method of the MessageFeedController that did not exist previously because GraphicOverlay layers are not loaded asynchronously.

dtedder:
related to #418 

Shared/ToolResourceProvider.h
Shared/ToolResourceProvider.cpp
- created a new signal for `sceneDoneLoading`
- connected the scene's done loading to emit the new signal

Shared/messages/MessageFeedsController.h
Shared/messages/MessageFeedsController.cpp
- swapped the member variable for `GeoView` to `Scene`
- added slot for `setScene`
- removed the `setGeoView` method
- connect to `ToolResourceProvider::sceneDoneLoading` to `MessageFeedsController::setScene`
- guard any calls to `setupFeed` with a check to the `m_scene` pointer
- remove code no longer needed in `setupFeeds`for getting the scene

Shared/packages/OpenMobileScenePackageController.cpp
- set to const to fix clazy warning about detached container

### DIFF
--- a/Shared/ToolResourceProvider.h
+++ b/Shared/ToolResourceProvider.h
@@ -81,6 +81,7 @@ public slots:
 signals:
   void sceneChanged();
   void geoViewChanged();
+  void sceneDoneLoading();
   void mapChanged();
   void spatialReferenceChanged();
   void setMouseCursorRequested(const QCursor& cursor);
--- a/Shared/ToolResourceProvider.h
+++ b/Shared/ToolResourceProvider.h
@@ -81,7 +81,6 @@ public slots:
 signals:
   void sceneChanged();
   void geoViewChanged();
-  void sceneDoneLoading();
   void mapChanged();
   void spatialReferenceChanged();
   void setMouseCursorRequested(const QCursor& cursor);
--- a/Shared/messages/MessageFeedsController.cpp
+++ b/Shared/messages/MessageFeedsController.cpp
@@ -67,10 +67,7 @@ MessageFeedsController::MessageFeedsController(QObject* parent) :
   m_messageFeeds(new MessageFeedListModel(this)),
   m_locationBroadcast(new LocationBroadcast(this))
 {
-  connect(ToolResourceProvider::instance(), &ToolResourceProvider::geoViewChanged, this, [this]
-  {
-    setGeoView(ToolResourceProvider::instance()->geoView());
-  });
+  connect(ToolResourceProvider::instance(), &ToolResourceProvider::sceneDoneLoading, this, &MessageFeedsController::setScene);
 
   ToolManager::instance().addTool(this);
 }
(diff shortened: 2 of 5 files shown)


Shared/ToolResourceProvider.h
```diff
@@ -81,6 +81,7 @@ public slots:
 signals:
   void sceneChanged();
   void geoViewChanged();
+  void sceneDoneLoading();
```
JamesMBallard: This does not seem like it should be a concern or a feature of the ToolResourceProvider.

Shared/messages/MessageFeedsController.cpp
```diff
@@ -85,13 +82,22 @@ MessageFeedsController::~MessageFeedsController()
 /*!
   \brief Sets the GeoView for the MessagesOverlay objects to \a geoView.
  */
-void MessageFeedsController::setGeoView(GeoView* geoView)
+void MessageFeedsController::setScene()
```
JamesMBallard: I would consider a different approach for this solution. It does appear there is an issue here if we start setting things up before scene is available, but this class is where that concern exists, so this class should be the one that can solve it.

There is a `sceneChanged` signal on the SceneQuickView you can listen for. Would that suffice here if there is no scene available yet?

dtedder: SceneQuickView::sceneChanged is what I was looking for all along. Previously, I was attempting to get a reference to the `-arcGISScene()` in the `ToolResourceManager::sceneChanged` signal. This isn't the same of course so when a package was used to set the scene instead of the default scene it's ends up still null when the slot on MessageFeedController was called.

Shared/messages/MessageFeedsController.h
```diff
@@ -83,6 +81,9 @@ class MessageFeedsController : public AbstractTool
 
   static Esri::ArcGISRuntime::SurfacePlacement toSurfacePlacement(const QString& surfacePlacement);
 
+private slots:
+  void setScene();
```
JamesMBallard: A few things here. A bit pedantic, but we don't really need to specify "slots" anymore. They're just functions that we can connect directly to signals as you are doing. Optional.

Also I would consider a different name since it's unclear what is driving this call. Maybe `setSceneFromGeoview_();` or `onGeoviewChanged_();`

trailing underscore optional.

JamesMBallard (commented):
A few thoughts on naming, but otherwise looks good.

Shared/messages/MessageFeedsController.h
dtedder: Pushed a new commit. removed the slots designation and just moved to be a private method. gave it a better name.

JamesMBallard (approved):

dtedder:
redirecting this PR to the new issue created specifically for the crash on startup #450. Investigation into why the layers are not updating specifically on physical devices will continue in another branch.

PR #447, opened 2025-05-23
Fix for Follow option missing for DynamicEntities/MessageFeeds

### LINKED ISSUE #441: No Follow option in context menu on a message feed
When press and hold on a graphic (e.g. from a message feed), the context menu should be shown and include the `Follow` option. The app should begin following the graphic when clicked on it

**Platforms**: Android, Windows

**Applications**: Vehicle, Handheld

dtedder:
combine all IdentifyResults into one property
simplify the assigning of contextmenu options in the identify results. search for points only until more than 1 GeoElement of type point is found. fix clazy warnings for non-const containers in range based for-loops

### DIFF
--- a/Shared/ContextMenuController.cpp
+++ b/Shared/ContextMenuController.cpp
@@ -133,13 +133,9 @@ void ContextMenuController::onMousePressedAndHeld(QMouseEvent& event)
     return;
 
   clearOptions();
-  for (const auto& feats : std::as_const(m_contextFeatures))
-    qDeleteAll(feats);
-  m_contextFeatures.clear();
-
-  for (const auto& graphics : std::as_const(m_contextGraphics))
-    qDeleteAll(graphics);
-  m_contextGraphics.clear();
+  for (const auto& geoElements : std::as_const(m_contextGeoElements))
+    qDeleteAll(geoElements);
+  m_contextGeoElements.clear();
 
   GeoView* geoView = ToolResourceProvider::instance()->geoView();
   if (!geoView)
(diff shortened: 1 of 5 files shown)


Shared/ContextMenuController.cpp
```diff
@@ -263,43 +259,33 @@ void ContextMenuController::setResultTitle(const QString& resultTitle)
...
-  if (m_contextFeatures.isEmpty() && m_contextGraphics.isEmpty())
+  if (m_contextGeoElements.isEmpty())
     return;
 
   // if we have at least 1 GeoElement, we can identify
   addOption(IDENTIFY_OPTION);
 
-  int pointGraphicsCount = 0;
-  for (const auto& geoElements : std::as_const(m_contextGraphics))
+  quint8 pointGeoElementCount = 0;
```
JamesMBallard, line 268: Could there be more than 255 geoelements? Probably not, but the size savings may not be worth the risk of an overflow in an edge case.

Shared/ContextMenuController.cpp
```diff
@@ -263,43 +259,33 @@ void ContextMenuController::setResultTitle(const QString& resultTitle)
...
+  for (const auto& geoElements : std::as_const(m_contextGeoElements))
   {
     for (const auto* geoElement : geoElements)
     {
       if (geoElement->geometry().geometryType() == GeometryType::Point)
-        pointGraphicsCount++;
+        if (++pointGeoElementCount > 1)
+          break;
```
JamesMBallard: Since there are nested conditions, I recommend using braces so it's clearer.

Shared/ContextMenuController.cpp
dtedder: There actually could be more than 255 GeoElements. But this variable will never need to store more than 2 as a value. The improvement here is actually exiting early once more than 1 point type is found. Maybe `size_t` or `qsizetype` is better? We just never need to continue checking once 2 are found since that takes 'FOLLOW' out as an available option.

JamesMBallard: quint8 makes perfect sense than, thanks for explaining. You could go with `auto` even, but your code is correct and appropriate.

dtedder:
@JamesMBallard I pushed up new commit to this branch that will disable 'follow' mode whenever the Navigation UI's Home button is clicked.

Shared/NavigationController.cpp
```diff
@@ -113,11 +114,14 @@ void NavigationController::updateGeoView()
...
-  if (m_is3d)
-  {
-    m_sceneView->setViewpointAsync(m_sceneView->arcGISScene()->initialViewpoint(), 1.0f);
-  }
+  if (!m_is3d)
+    return;
+
+  // stop following if the home button is clicked
+  if (auto* followPositionController = ToolManager::instance().tool<FollowPositionController>(); followPositionController)
+    followPositionController->setFollowMode(FollowPositionController::FollowMode::Disabled);
```
JamesMBallard: Can this be moved to its own function and called by whatever is calling `zoomToInitialLocation`? This sounds like the desirable behavior, but it would be a side effect of calling `zoomToInitialLocation`, since that name doesn't mention anything about changing the follow mode.

dtedder: Good point. I found a link to the `FollowHud` in the QML that has a method to stop following. I undid the cpp change and called this instead.

JamesMBallard (approved):

PR #451, opened 2025-06-09
Check for selected target in Alert wizard

### LINKED ISSUE #391: Layer filter in Condition wizard not applied
The UI for the Alerts wizard presents users with a drop-down for selecting a Layer. However the Identify action will only ever return a single entity and does not include any means to identify the selection action on the layer that was selected. This needs to consider the layer selected in the drop-down.

dtedder:
added a property for tracking the selected target in the alert create wizard for spatial
created a method to get the single element returned in an identify results geoElements list added logic to check for the selected target in the click event of AlertConditionController added reset to -1 for the selected target index to the qml

related to #391

### DIFF
--- a/Shared/alerts/AlertConditionsController.cpp
+++ b/Shared/alerts/AlertConditionsController.cpp
@@ -715,16 +715,16 @@ const std::optional<qint64> AlertConditionsController::getPickedElementId(const 
   // check again after the completion of the identify and abort if tool
   // is no longer the active tool or there were no geoelements returned
   if (!isActive())
-    return {};
-  if (geoElements.count() < 1)
-    return {};
+    return std::nullopt;
+  if (geoElements.isEmpty())
+    return std::nullopt;
 
   // get the entity id for dynamic entity types
   auto* geoElement = geoElements.first();
   if (auto* observation = dynamic_cast<DynamicEntityObservation*>(geoElement); observation)
   {
     observation->deleteLater();
-    return observation->dynamicEntity()->entityId();
+    return std::make_optional<qint64>(observation->dynamicEntity()->entityId());
   }
 
   // check the attribute table if feature type
@@ -733,17 +733,20 @@ const std::optional<qint64> AlertConditionsController::getPickedElementId(const 
     const auto* attributes = feature->attributes();
     auto* table = feature->featureTable();
     if (!attributes)
-      return {};
+      return std::nullopt;
     if (!table)
-      return {};
+      return std::nullopt;
 
     const auto primaryKeyField = primaryKeyFieldName(table);
     if (primaryKeyField.isEmpty())
-      return {};
+      return std::nullopt;
     if (!attributes->containsAttribute(primaryKeyField))
-      return {};
+      return std::nullopt;
 
-    return attributes->attributeValue(primaryKeyField).toInt();
+    if (auto primaryKeyValue = attributes->attributeValue(primaryKeyField); primaryKeyValue.canConvert<int>())
+      return attributes->attributeValue(primaryKeyField).toInt();
+    else
+      return std::nullopt;
   }
 
   // graphics, just use the index of the graphic in the layer. not sure we have a case where this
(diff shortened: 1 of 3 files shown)


Shared/alerts/AlertConditionsController.h
```diff
@@ -109,6 +110,7 @@ private slots:
   QJsonObject conditionToJson(AlertCondition* condition) const;
   QFuture<bool> addConditionFromJson(const QJsonObject& json);
   void addStoredConditions();
+  const std::optional<qint64> getPickedElementId(const QList<Esri::ArcGISRuntime::GeoElement*>& geoElements) const;
```
JamesMBallard, line 114: Nice, clean API 👍

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -710,6 +710,58 @@ void AlertConditionsController::onLayersChanged()
   addStoredConditions();
 }
 
+const std::optional<qint64> AlertConditionsController::getPickedElementId(const QList<GeoElement*>& geoElements) const
+{
+  // check again after the completion of the identify and abort if tool
+  // is no longer the active tool or there were no geoelements returned
+  if (!isActive())
+    return {};
+  if (geoElements.count() < 1)
```
JamesMBallard: isEmpty is typical idiomatic Qt for checking containers for emptiness.

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -710,6 +710,58 @@ void AlertConditionsController::onLayersChanged()
...
+    if (!table)
+      return {};
+
+    const auto primaryKeyField = primaryKeyFieldName(table);
+    if (primaryKeyField.isEmpty())
+      return {};
+    if (!attributes->containsAttribute(primaryKeyField))
+      return {};
+
+    return attributes->attributeValue(primaryKeyField).toInt();
```
JamesMBallard: On all code paths where you return an actual value, I think `std::make_optional` or just `std::optional{ ... }` is more explicit. My two cents would be to add something like that to make it clearer to the reader that's what is really happening. Without it, it is an implicit conversion.

JamesMBallard: Are there any warnings related to converting `toInt` to `qint64`?

Shared/alerts/AlertConditionsController.h
```diff
@@ -144,6 +146,7 @@ private slots:
   QHash<QString,QString> m_messageFeedTypesToNames;
 
   QMetaObject::Connection m_mouseClickConnection;
+  qint16 m_selectedTargetNameIndex = -1;
```
JamesMBallard: If this is being passed in from QML land (I see the method is `Q_INVOKABLE`), then I would consider another data type. 

These are your choices, so I'd probably go with int.
https://doc.qt.io/archives/qt-5.15/qtqml-typesystem-basictypes.html#basic-types-provided-by-the-qml-language

QML has no support for `qint16` so you're truncating info from the caller, which is probably fine but not worth dealing with any unintended consequences (which would probably not happen, but why risk it).

Shared/alerts/AlertConditionsController.cpp
dtedder: Yep, will swap this out.

Shared/alerts/AlertConditionsController.cpp
dtedder: Agreed, I'll make the update for the returns. Maybe also `std::nullopt` instead of the `{}` as well when there's nothing to return?

dtedder: Was not getting a warning here for any type conversion. However I am pretty sure this should be protected with `canConvert` to make sure it's not null from the attribute table.

Shared/alerts/AlertConditionsController.h
dtedder: Understood. Will switch to just `int`.

Shared/alerts/AlertConditionsController.cpp
JamesMBallard: >Maybe also std::nullopt

I will leave that up to you. IMO that's more subjective. Whenever you return `{}`, the reader can't jump to conclusions about what the type is, so it's less confusing. I do think returning `std::nullopt` is better overall.

JamesMBallard (approved):

dtedder:
@JamesMBallard I pushed another commit just now. Nothing big, just forgot to also update the type to plain `int` on the invokable method decl/def.

1f2f485618099fb1a4eedaf18d8f65a03f759382

dtedder:
@JamesMBallard another small cleanup based on testing. Clicking on the map where nothing exists will de-select the layer the user picked in the drop-down. Also, switching target layers should clear any previous ID to invalidate the form so the user can't click next and mistakenly apply the ID to a different target layer.

JamesMBallard (approved):
Latest changes 👍

PR #452, opened 2025-06-11
Use model properties for selected basemap

### LINKED ISSUE #415: Basemaps panel doesn't highlight the basemap that is being displayed upon startup
## Description
When you run the DSA app without changing any settings and using the default demo data, the Topographic basemap is the default in the AppConfig file, and it is shown in the scene by default.  However, the Basemaps panel highlights the Imagery basemap in the Basemaps panel. (perhaps it's just highlighting the first baseman by default?)

Image

## Expected Behavior
The basemap that is drawn on the scene should be the one highlighted in the Basemaps panel. 

## Test steps:
- Open DSA with the Default configuration (so the Topographic basemap is already downloaded and loaded into the scene)
- Tap the Basemap tool
  - [ ]  The basemap that is highlighted should be the basemap that is already displayed in the scene (Topographic)
- Select the imagery basemap and close the Basemap tool
- Tap the Basemap tool
  - [ ]  The basemap that is highlighted should be the basemap that is already displayed in the scene (Imagery)
- [ ] In general, there is never a consistency issue where the basemap that is highlighted in blue is not the same as the basemap being displayed on the scene, as you switch basemaps, open/close the Basemap tool, open/close the app.

dtedder:
Existing GridView was using hover mouse event to switch the currentIndex which was leading to the unexpected behavior in the associated issue. The currentIndex is now bound to the selected basemap's index which is updated on the controller on mouse click. There is also a new property for the path (value updated along with index) that is bound in the footer text element instead of the QML property that was used before.

### DIFF
--- a/Shared/BasemapPickerController.h
+++ b/Shared/BasemapPickerController.h
@@ -82,7 +82,7 @@ private:
   QString             m_basemapDataPath;
   QString             m_defaultBasemap = "topographic";
   int                 m_selectedBasemapIndex = -1;
-  QString             m_selectedBasemapPath = "";
+  QString             m_selectedBasemapPath;
 
 private:
   void findBasemaps();
--- a/Shared/BasemapPickerController.h
+++ b/Shared/BasemapPickerController.h
@@ -41,6 +41,8 @@ class BasemapPickerController : public AbstractTool
   Q_OBJECT
 
   Q_PROPERTY(QAbstractListModel* tileCacheModel READ tileCacheModel NOTIFY tileCacheModelChanged)
+  Q_PROPERTY(int selectedBasemapIndex READ selectedBasemapIndex NOTIFY selectedBasemapIndexChanged)
+  Q_PROPERTY(QString selectedBasemapPath READ selectedBasemapPath NOTIFY selectedBasemapIndexChanged)
 
 public:
   static const QString DEFAULT_BASEMAP_PROPERTYNAME;
@@ -61,6 +63,8 @@ public:
   void setBasemapDataPath(const QString& dataPath);
   QString defaultBasemap() const { return m_defaultBasemap; }
   void setDefaultBasemap(const QString& defaultBasemap);
+  int selectedBasemapIndex() const { return m_selectedBasemapIndex; }
+  QString selectedBasemapPath() const { return m_selectedBasemapPath; }
 
 public slots:
   void onBasemapDataPathChanged();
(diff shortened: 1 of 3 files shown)


Shared/BasemapPickerController.h
```diff
@@ -70,12 +74,15 @@ public slots:
...
   void toolErrorOccurred(const QString& errorMessage, const QString& additionalMessage);
+  void selectedBasemapIndexChanged();
 
 private:
   TileCacheListModel* m_tileCacheModel;
   int                 m_defaultBasemapIndex = 0;
   QString             m_basemapDataPath;
   QString             m_defaultBasemap = "topographic";
+  int                 m_selectedBasemapIndex = -1;
+  QString             m_selectedBasemapPath = "";
```
JamesMBallard:
```suggestion
  QString             m_selectedBasemapPath;
```
Default constructor does the same thing, perhaps more efficiently.

Shared/qml/BasemapPicker.qml
```diff
@@ -76,27 +76,20 @@ DsaPanel {
...
                 onClicked: {
-                    toolController.basemapSelected(index);
-                    basemapSelected();
-                }
-                onHoveredChanged: {
-                    if (containsMouse) {
-                        basemapsList.currentIndex = index
-                        basemapsList.currentPath = path
+                    if (index !== toolController.selectedBasemapIndex)
+                    {
```
JamesMBallard:
```suggestion
                    if (index !== toolController.selectedBasemapIndex) {
```

Shared/BasemapPickerController.h
```diff
@@ -70,12 +74,15 @@ public slots:
...
   void basemapChanged(Esri::ArcGISRuntime::Basemap* basemap, QString name = "");
   void toolErrorOccurred(const QString& errorMessage, const QString& additionalMessage);
+  void selectedBasemapIndexChanged();
 
 private:
   TileCacheListModel* m_tileCacheModel;
   int                 m_defaultBasemapIndex = 0;
   QString             m_basemapDataPath;
   QString             m_defaultBasemap = "topographic";
+  int                 m_selectedBasemapIndex = -1;
```
JamesMBallard, line 84: Any concerns -1 could be read and used somewhere?

Shared/qml/BasemapPicker.qml
```diff
@@ -110,6 +103,7 @@ DsaPanel {
         cellWidth: 128 * scaleFactor
         cellHeight: 128 * scaleFactor
         width: 2 * cellWidth
+        currentIndex: toolController.selectedBasemapIndex
```
JamesMBallard, line 105: Presumably this will get initialized to -1 once loaded. Is that an issue?

Shared/BasemapPickerController.h
dtedder: I don't think -1 is a concern. It is only ever ready by the QML. -1 is the default value for a Combobox's `currentIndex` property when there are no elements. QML's combobox seems to follow the same pattern as other UI controls I've worked with in WPF, WinForms, HTML. -1 is used to indicate nothing is selected.

Shared/qml/BasemapPicker.qml
dtedder: Same as above. This was actually my intent. Even if the data folder selected contains nothing, -1 will still be a valid value for this component. Once the initialization happens in the app, the property will be updated to match what's in the config and the UI will be notified.

JamesMBallard (approved):

PR #454, opened 2025-06-18
Adding build scripts

har13205:
PR to add build scripts for windows and android

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,35 @@
+# Build Instructions
+
+
+1. Modify the following variables in the script based on your local setup:
+
+- QT_ROOT : Path to your Qt installation
+- ANDROID_NDK_ROOT : Path to your Android NDK
+- ANDROID_PLATFORM : Android API level (e.g., `android-34`)
+
+
+2. Clone the Dynamic Situational Awareness (DSA) repository
+
+```
+git clone https://github.com/Esri/dynamic-situational-awareness-qt.git
+```
+
+3. Clone the ArcGIS Maps SDK for Qt Toolkit repository
+
+```
+cd dynamic-situational-awareness-qt
+git clone https://github.com/Esri/arcgis-maps-sdk-toolkit-qt.git
+```
+
+4. Navigate to the scripts directory and run the appropriate build script, specifying the application type (Vehicle or Handheld):
+
+- For windows:
+```
+build_windows.bat <application>
+```
+
+- For android:
+
+```
+build_android.bat <application>
+```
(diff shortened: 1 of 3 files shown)


scripts/build_android.bat
```diff
@@ -0,0 +1,80 @@
...
+    echo [ERROR] Application name argument missing.
+    exit /b 1
+)
+
+set "QT_ROOT=C:\Qt6.8.2\6.8.2"
+set "QT_ANDROID_BIN=%QT_ROOT%\android_arm64_v8a\bin"
+set "QT_DEPLOY_TOOL=%QT_ROOT%\mingw_64\bin\androiddeployqt.exe"
+set "MAKE_EXE=%USERPROFILE%\AppData\Local\Android\Sdk\ndk\28.0.12433566\prebuilt\windows-x86_64\bin\make.exe"
+set "ANDROID_NDK_ROOT=%USERPROFILE%\AppData\Local\Android\Sdk\ndk\28.0.12433566"
+set "ANDROID_PLATFORM=android-34"
```
JamesMBallard: There are some assumptions here. Unclear if this works for everyone, or if the DSA team has standardized around these paths.

scripts/build_android.bat
```diff
@@ -0,0 +1,80 @@
...
+cd "%BUILD_DIR%" || exit /b 1
+
+:: =============================
+:: Run QMake and Make
+:: =============================
+echo [INFO] Running qmake...
+call "%QT_ANDROID_BIN%\qmake.bat" ^
+    "%PRO_FILE%" ^
+    -spec android-clang ^
+    CONFIG+=debug CONFIG+=qml_debug || exit /b 1
```
JamesMBallard, line 45: This will create a debug build. That simplifies things if you don't want to deal with codesigning, but it will be problematic if anyone is checking performance. I recommend having both options in here, and the release build mode should be the default. If that is too much scope creep, than only the release build should suffice.

kerryrobinson: We like to build debug so that the user can use Android Studio to copy their own data into the sandbox (this will be resolved in the next version of DSA). But it would be a good idea to add a flag to the script so the user can specify `debug` for the build if they want.

JamesMBallard: If a debug build is expected, that sounds fine. Presumably someone will eventually want to run a release build, so that should be an option but for now this seems sufficient.

scripts/README.md
```diff
@@ -0,0 +1,35 @@
+# Build Instructions
+
+
+1. Modify the following variables in the script based on your local setup:
+
+- QT_ROOT : Path to your Qt installation
+- ANDROID_NDK_ROOT : Path to your Android NDK
+- ANDROID_PLATFORM : Android API level (e.g., `android-34`)
```
JamesMBallard, line 8: This may be overkill at this point, but the script should allow these to be read in from the environment. Without that, everyone will have to hold local changes that will show up in git and need to remember to not accidentally commit it.

JamesMBallard (approved):
Recommendation for an improvement, but possibly out of scope for now. Consider if it's needed for the team, or circle back later if it becomes problematic without it.

dtedder (approved):
👍

PR #459, opened 2025-07-29
Add Grids

### LINKED ISSUE #421: Add Scene Grid support

dtedder:
added grid button to map navigation tool area. linked grid mode to coordinate format

#421

### DIFF
--- a/Shared/AbstractTool.cpp
+++ b/Shared/AbstractTool.cpp
@@ -16,10 +16,14 @@
 
 #include "AbstractTool.h"
 
+// STL headers
+#include <algorithm>
+
 using namespace Esri::ArcGISRuntime;
 
 namespace Dsa
 {
+
 /*!
   \class Dsa::AbstractTool
   \inmodule ArcGISQtToolkit
@@ -71,9 +75,18 @@ bool AbstractTool::handleClick(const Point& pos)
   and the value is a QVariant. Properties are useful for general settings
   which you wish to persist for the tool, such as default modes.
  */
-void AbstractTool::setProperties(const QVariantMap&)
+void AbstractTool::toolInitProperties(const QVariantMap&)
 {
+}
 
+/*
+ * Derived tools should override this method to implement logic that
+ * will determine whether or not the tool's setProperties method should
+ * be invoked again to re-initialize the tool when a property changes.
+ */
+bool AbstractTool::shouldSetProperties(const QString&)
+{
+  return false;
 }
 
 /*!
@@ -97,6 +110,11 @@ void AbstractTool::setActive(bool active)
 bool AbstractTool::isActive() const
 {
   return m_active;
+}
+
+bool AbstractTool::setContainsString(const std::unordered_set<QString>& strSet, const QString& str)
+{
+  return std::any_of(std::cbegin(strSet), std::end(strSet), [str](const QString& p) { return str == p; });
 }
 
 // Signals
--- a/Shared/AbstractTool.cpp
+++ b/Shared/AbstractTool.cpp
@@ -114,7 +114,7 @@ bool AbstractTool::isActive() const
 
 bool AbstractTool::setContainsString(const std::unordered_set<QString>& strSet, const QString& str)
 {
-  return std::any_of(std::cbegin(strSet), std::end(strSet), [str](const QString& p) { return str == p; });
+  return std::any_of(std::cbegin(strSet), std::cend(strSet), [str](const QString& p) { return str == p; });
 }
 
 // Signals
--- a/Shared/GridController.cpp
+++ b/Shared/GridController.cpp
@@ -33,9 +33,12 @@
 #include "USNGGrid.h"
 #include "UTMGrid.h"
 
+// Qt headers
+#include <QVariantMap>
+
 // STL headers
 #include <algorithm>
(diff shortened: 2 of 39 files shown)


Shared/GridController.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+  // stop listening for new tools being added
+  disconnect(m_conToolAdded);
+}
+
+void GridController::setProperties(const QVariantMap &properties)
+{
+  // create a string lookup to use lat long for everything but utm/mgrs/usng
+  using namespace Esri::ArcGISRuntime::Toolkit;
+  const static std::vector<QString> nonDefaultFormatNames{
```
JamesMBallard: Subjective, but we typically go with `static const`.

Shared/GridController.cpp
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  auto* geoView = ToolResourceProvider::instance()->geoView();
+  if (!geoView)
+    return;
+
+  // access tool properties from the config
+  auto formatLookup = QString{QStringLiteral("default")};
+  auto coordinateFormat = properties["CoordinateFormat"].toString();
+  if (std::find(nonDefaultFormatNames.cbegin(), nonDefaultFormatNames.cend(), coordinateFormat) != nonDefaultFormatNames.cend())
```
JamesMBallard: For three entries it hardly matters, but I would still recommend an unordered_set for quick lookups vs linear search through a vector.

Also there has been an emphasis on using the free functions std::begin, std::cbegin, etc over the members. I go back and forth but I tend to stick with the free functions.

Shared/GridController.h
```diff
@@ -0,0 +1,59 @@
...
+
+#ifndef GRIDCONTROLLER_H
+#define GRIDCONTROLLER_H
+
+// DSA headers
+#include "AbstractTool.h"
+
+// Qt headers
+#include <QString>
+#include <QVariantMap>
```
JamesMBallard: You can probably forward declare this one. QtContainerFwd should have all the forwards declares to make that easy. No need to copy paste out of that file, just include it here. Then you can include `<QVariantMap>` over in the .cpp file.

Shared/NavigationController.h
```diff
@@ -50,6 +50,7 @@ class NavigationController : public AbstractTool
   Q_INVOKABLE void pan();
   Q_INVOKABLE void setRotation();
   Q_INVOKABLE void set2D();
+  Q_INVOKABLE void setShowGrid(bool show);
```
JamesMBallard: This doesn't make much difference, but I would think `showGrid` would make sense as a property in the `Q_PROPERTY`. Invokable methods are typically more meaningful operations and properties are more like simple state management stuff, and this feels more like simple state management.

I recommend making this a property. Maybe:
```c++
Q_PROPERTY(bool showGrid READ showGrid WRITE setShowGrid NOTIFY showGridChanged)
```

Then `void showGridChanged(bool show);` wouldn't need the value passed on the signal. That should be simple enough to change elsewhere where it's used.

If you want to address these separately, that works as well. For example if this makes sense as a property, but you need to keep passing the value on the signal.

Shared/qml/NavigationTool.qml
```diff
@@ -32,6 +32,10 @@ Item {
     width: controlsColumn.width + 10 * scaleFactor
     property alias radius: backgroundRecatangle.radius
 
+    GridController {
+        id: gridController
+    }
```
JamesMBallard: If the `showGrid` logic is a property, you can bind it here as long as that button has an id.
like
```qml
    GridController {
        id: gridController
        showGrid: overlayButton.selected
    }
```

That's another benefit over invokable methods.

Shared/NavigationController.h
dtedder: Moved this to use Q_PROPERTY and moving everything over to GridController directly made things much simpler.

Shared/GridController.cpp
dtedder: switched to free functions and unordered_set

Shared/GridController.h
dtedder: 👍

Shared/qml/NavigationTool.qml
dtedder: 👍

Shared/GridController.cpp
dtedder: 👍

JamesMBallard (approved):

Shared/AbstractTool.cpp
```diff
@@ -99,6 +112,11 @@ bool AbstractTool::isActive() const
   return m_active;
 }
 
+bool AbstractTool::setContainsString(const std::unordered_set<QString>& strSet, const QString& str)
+{
+  return std::any_of(std::cbegin(strSet), std::end(strSet), [str](const QString& p) { return str == p; });
```
JamesMBallard:
```suggestion
  return std::any_of(std::cbegin(strSet), std::cend(strSet), [str](const QString& p) { return str == p; });
```

JamesMBallard (approved):
Very minor nitpick, but very good overall. I like the new helper and approach.  No need for another review.

JamesMBallard (approved):

PR #460, opened 2025-08-08
enable user custom configuration download from portal, web or local file

dtedder:
- switched to using a different approach to downloading and unzipping in ConfigurationController that is more suitable to managing multiple configurations instead of just the default
- added a new resource icon for browsing files
- fixed a bug that failed to find data in the config folder due to the folder separator being on the end of the root directory
- updated the gradle dependencies to satisfy warning from the build tools
- included concurrent for running zip extraction async
- removed multimedia from DSA for issues with iOS builds
- renamed properties in the config file to be consistent in AppConstants


Shared/AppConstants.h
```diff
@@ -24,15 +24,34 @@ namespace Dsa {
...
+  inline static const QString PROPERTYNAME_UNIT_OF_MEASUREMENT = QStringLiteral("UnitOfMeasurement");
   inline static const QString UNIT_METERS = QStringLiteral("meters");
   inline static const QString UNIT_FEET = QStringLiteral("feet");
-  inline static const QString LAYERS_PROPERTYNAME = QStringLiteral("Layers");
-  inline static const QString CURRENTSCENE_PROPERTYNAME = QStringLiteral("CurrentPackage");
-  inline static const QString SCENEINDEX_PROPERTYNAME = QStringLiteral("SceneIndex");
-  inline static const QString INITIALLOCATION_PROPERTYNAME = QStringLiteral("InitialLocation");
-  inline static const QString LAYER_NAME_SCENEVIEW_LOCATION = QStringLiteral("SceneViewLocationOverlay");
+
+  inline static const QString BOOL_TRUE = QStringLiteral("true");
```
JamesMBallard, line 54: Is this used anywhere?

Shared/configurations/Configuration.h
```diff
@@ -19,12 +19,14 @@
...
 // Qt headers
 #include <QString>
+#include <QUrl>
 
 namespace Dsa {
 
 class Configuration
 {
 public:
+  Configuration() = default;
```
JamesMBallard: Implementation should go into .cpp files, even `= default`.

Shared/configurations/Configuration.h
```diff
@@ -49,13 +60,21 @@ class Configuration
...
   QString m_name;
-  QString m_url;
+  QString m_urlStr;
+  QUrl m_url;
   bool m_selected = false;
   bool m_loaded = false;
+  bool m_isCancellable = true;
   int m_percentDownloaded = 0;
+  bool m_downloadCancelled = false;
+  int m_percentExtracted = 0;
```
JamesMBallard, line 77: I am surprised to see `percentExtracted`. That will be nice for users to see, but I thought extracting is typically quick so any status bar would be barely noticeable. On the other hand if the zips are large, it might make sense.

What's the motivation for adding this?

Shared/configurations/Configuration.cpp
```diff
@@ -95,4 +142,32 @@ void Configuration::setPercentDownloaded(int percentDownloaded)
...
+}
+
+void Configuration::setPercentExtracted(int percentExtracted)
+{
+  m_percentExtracted = percentExtracted;
+}
+
+bool Configuration::inProgress() const
+{
+  auto progress = m_percentDownloaded + m_percentExtracted;
```
JamesMBallard:
```suggestion
  const auto progress = m_percentDownloaded + m_percentExtracted;
```

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,26 +79,183 @@ void ConfigurationController::select(int index)
...
 }
 
+bool removeDownloadedFile(const QString& downloadFilePath)
+{
+  // remove any remaining progress if request was aborted or after the zip has been extracted
+  QFile f{downloadFilePath};
+  if (f.exists())
+    return f.remove();
+
+  return false;
```
JamesMBallard: Should this return true? If there was no file to remove, I suppose it successfully removed nothing in that case.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,26 +79,183 @@ void ConfigurationController::select(int index)
...
+  // find the previous root data directory value that was configured
+  auto configObj = configDoc.object();
+  const auto rootDataDirectoryVar = configObj.value(AppConstants::PROPERTYNAME_ROOT_DATA_DIRECTORY);
+  if (rootDataDirectoryVar.isUndefined())
+    return false;
+
+  // strip off any trailing slashes if they exist
+  auto rootDataDirectoryCleaned = rootDataDirectoryVar.toString();
+  while (rootDataDirectoryCleaned.back() == '\\' || rootDataDirectoryCleaned.back() == '/')
+    rootDataDirectoryCleaned.chop(1);
```
JamesMBallard: This logic seems good, but you may get similar logic from either QDir or QFile that strip trailing slashes, or remove redundant ones. It may be overkill and this is better, but give it a look.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,26 +79,183 @@ void ConfigurationController::select(int index)
...
 {
-  // prevent calling the download function if something is already in progress
-  if (m_downloadInProgress)
+  // fetch a copy of the selected model item from the listview
+  const auto configuration = m_configurationListModel->at(index);
+  const auto configurationUrl = configuration.url();
+  const auto configurationName = configuration.name();
+  if (configurationUrl.isLocalFile())
+  {
+    m_configurationListModel->setDataByName(configurationName, 100, ConfigurationListModel::PercentDownloaded);
```
JamesMBallard, line 330: Just a passing comment, but I don't know what '100' means here.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,26 +79,183 @@ void ConfigurationController::select(int index)
...
     return;
+  }
 
-  // get the url from the selected item
-  auto i = m_configurationListModel->index(index, 0);
-  auto url = m_configurationListModel->data(i, ConfigurationListModel::Roles::Url);
-  m_activeDownloadIndex = index;
+  // attempt to download the data as a portal item if it matches the item id url pattern
+  const auto configurationUrlStr = configuration.urlStr();
+  static QRegularExpression regexPortalItem{REGEX_PORTAL_ITEM_URL};
```
JamesMBallard: Can this also be const?

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,26 +79,183 @@ void ConfigurationController::select(int index)
...
+      {
+        portal->deleteLater();
+        resetConfigurationDeviceStatus(configurationName);
+        removeDownloadedFile(downloadFilePath);
+        emit toolErrorOccurred(QString("Downloading the item failed: %1").arg(e.what()), QStringLiteral("Download Error"));
+      });
+      connect(portalItem, &PortalItem::fetchDataProgressChanged, this, [this, configurationName](const NetworkRequestProgress& progress)
+      {
+        m_configurationListModel->setDataByName(configurationName, progress.progressPercentage(), ConfigurationListModel::PercentDownloaded);
+      });
```
JamesMBallard: I would move this connection to before the `fetchDataAsync` call is made.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,26 +79,183 @@ void ConfigurationController::select(int index)
...
-  m_activeDownloadIndex = index;
+  // attempt to download the data as a portal item if it matches the item id url pattern
+  const auto configurationUrlStr = configuration.urlStr();
+  static QRegularExpression regexPortalItem{REGEX_PORTAL_ITEM_URL};
+  if (regexPortalItem.match(configurationUrlStr).hasMatch())
+  {
+    PortalItem checkPortalItem{configurationUrl}; // using a local portal item here to parse the portal and item id out of the url
+    auto* portal = new Portal(checkPortalItem.portal()->url(), this);
+    auto* portalItem = new PortalItem(portal, checkPortalItem.itemId(), portal);
+    connect(portalItem, &PortalItem::doneLoading, this, [this, portal, portalItem, configurationName](const Error& error)
```
JamesMBallard, line 342: This is optional, but if there are large lambdas, consider splitting them out into their own methods. For example `onPortalItemLoaded_` and then this connection can just call that directly.

The `download` method looks a bit large with several connection inside of it.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -95,46 +267,75 @@ void ConfigurationController::download(int index)
...
+  const auto& configuration = m_configurationListModel->at(index);
+  QDir configurationDirectory{m_configurationsDirectory.filePath(configuration.name())};
+  if (configurationDirectory.exists())
+  {
+    configurationDirectory.removeRecursively();
+    resetConfigurationDeviceStatus(configuration.name());
+  }
+
+  // if user also requested to remove the configuration from the list
+  // remove it's corresponding model and update the list model
```
JamesMBallard:
```suggestion
  // remove its corresponding model and update the list model
```
Hyper nitpick

Shared/configurations/ConfigurationListModel.cpp
```diff
@@ -46,11 +46,29 @@ void ConfigurationListModel::cancel(int index)
...
+
+Configuration ConfigurationListModel::byName(const QString& name) const
+{
+  for (qsizetype i = 0; i < m_configurations.count(); i++)
+  {
+    if (m_configurations[i].name() == name)
+      return m_configurations[i];
+  }
+  return Configuration{};
+}
```
JamesMBallard: This is a good candidate for using std::find_if.

Shared/configurations/ConfigurationListModel.cpp
```diff
@@ -109,12 +155,28 @@ bool ConfigurationListModel::setData(const QModelIndex& index, const QVariant& v
...
 
   emit dataChanged(index, index);
   return true;
 }
 
+bool ConfigurationListModel::setDataByName(const QString& name, const QVariant& value, int role)
+{
+  for (qsizetype i = 0; i < m_configurations.count(); i++)
+  {
+    if (m_configurations[i].name() == name)
```
JamesMBallard, line 172: Lots of linear searches through `m_configurations`. Would it make more sense as a different data structure like a hash table indexed by name? Probably not since this is a list model and needs to be arbitrarily ordered.

Shared/utilities/ZipHelper.h
```diff
@@ -106,12 +107,15 @@ class ZipHelper : public QObject
...
   bool zrOpen();
+  qreal percentTotal();
   const QString& path() const { return m_Path; }
 
 private:
   QString                     m_Path;
   unzFile                     m_unzFile;
   zipFile                     m_zipFile;
+  quint64 m_bytesExtracted;
+  quint64 m_bytesTotalUncompressed;
```
JamesMBallard: These should be defaulted to 0 or something.

Shared/qml/Options.qml
```diff
@@ -438,4 +480,21 @@ Rectangle {
             optionsRoot.visible = false;
         }
     }
+
+    DsaYesNoCancelDialog {
+        id: confirmRemoveConfigurationDialog
+        onAccepted: removeConfiguration()
+        onRejected: removeConfiguration()
```
JamesMBallard: Do the same thing on both paths? Doesn't seem right.

Shared/AppConstants.h
dtedder: https://github.com/Esri/dynamic-situational-awareness-qt/blob/dtedder/402_download_custom_configurations/Shared/DsaController.cpp#L516C83-L516C92
https://github.com/Esri/dynamic-situational-awareness-qt/blob/dtedder/402_download_custom_configurations/Shared/DsaController.cpp#L510C86-L510C95

Shared/configurations/Configuration.h
dtedder: The size of the Monterey CA data that we provide is large enough to cause the UI to lockup long enough for it to appear that something has gone wrong and for the user to cancel or close the app leaving it in an undefined state for the configuration.  The ZipHelper class has been modified to include an _overall_ percent complete. This was added along with the concurrent running of `extractAll` to keep from blocking the UI while providing feedback in the progress bar so that it is clear the app is still performing work.

JamesMBallard: Cool, sounds good then.

Shared/qml/Options.qml
dtedder: I made a change recently that I think might remove the need for this extra dialog altogether. But it was used in order to capture the difference between No and Cancel on the Yes/No/Cancel prompt. The built-in Qt dialog just gives 'rejected' signal for both. Here the custom dialog stores the state as a property so it can be checked in `removeConfiguration()`.

I will re-evaluate this section of the PR and  maybe push a commit to remove some of the unnecessary parts.

Shared/configurations/ConfigurationListModel.cpp
dtedder: QList of pointers is used in almost every case in our code. I have not created a custom list myself yet so I decided to stick with this as the strategy. There are cases where the index value is used for creating the QModelIndex. I will switch any cases to `std::find_if` that are appropriate.

Shared/configurations/ConfigurationController.cpp
dtedder: Good catch. Didn't realize I had this after the call to fetch.

Shared/configurations/ConfigurationController.cpp
dtedder: Maybe a templated setter like this makes it a little more readable?
```suggestion
    m_configurationListModel->set<ConfigurationListModel::PercentDownloaded>(configurationName, 100);
```

Shared/configurations/ConfigurationController.cpp
dtedder: I tested out a few different approaches. `QDir::cleanPath()` seemed like the best option. However, I think we actually need to check explicitly for both `\\` as well as `/` in this unique case. If the user prepares a configuration file using Windows paths and later downloads it to any other platform, the `QDir::cleanPath()` method executing on the target system actually does not strip off the `\\` because that's not the file separator on POSIX compliant systems.

Shared/configurations/ConfigurationController.cpp
dtedder: Taking another look, I never ended up needing this result anywhere in the code that makes use of it. I will switch to void.

Shared/configurations/ConfigurationController.cpp
JamesMBallard: That's a bit better, but I understand the meaning now after you've explained it. I think it's ok as-is, but primarily because that template stuff looks like overkill.

Shared/configurations/ConfigurationListModel.cpp
dtedder: Pushed up a commit for this one. I think most of the other cases it might not be possible to use since the index of the item in the list is necessary.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,77 +79,308 @@ void ConfigurationController::select(int index)
...
+      quint64 bytesWritten = 0;
+      quint8 percentTotal = 0;
+      constexpr qint64 sizeLine = 1024 * 16;
+
+      // write the lines to the temporary 'download' file and update the
+      // percentage downloaded by whole integers only to limit updates to the UI
+      while (!fileContent.atEnd())
+      {
+        bytesWritten += fileDownload.write(fileContent.readLine(sizeLine));
+        const quint8 percentLine = bytesWritten * 1.0 / bytesTotal * 100.0;
```
JamesMBallard, line 312:
```suggestion
        const quint8 percentLine = bytesWritten / bytesTotal * 100.0;
```
I assume '* 1' can be removed.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,77 +79,308 @@ void ConfigurationController::select(int index)
...
+      fileContent.open(QIODevice::ReadOnly);
+
+      QFile fileDownload{contentDownloadStr};
+      fileDownload.open(QIODevice::WriteOnly);
+      quint64 bytesWritten = 0;
+      quint8 percentTotal = 0;
+      constexpr qint64 sizeLine = 1024 * 16;
+
+      // write the lines to the temporary 'download' file and update the
+      // percentage downloaded by whole integers only to limit updates to the UI
```
JamesMBallard, line 308: Can you help me understand what's happening here? We have all the bytes, but we're artificially slowing it down so we can update the UI progress bar, is that right?

Shared/configurations/ConfigurationController.cpp
```diff
@@ -154,6 +400,27 @@ void ConfigurationController::storeConfigurations()
...
 
+bool ConfigurationController::isDownloadCancelled(const QString& configurationName)
+{
+  return m_configurationListModel->dataByName(configurationName, ConfigurationListModel::DownloadCancelled).toBool();
+}
+
+bool ConfigurationController::deviceHasRoomForDownload(qint64 bytesToDownload)
+{
+  QStorageInfo storageInfo(m_downloadFolder);
+  auto bytesAvailable = storageInfo.bytesAvailable() * 0.95;
```
JamesMBallard: Optional, but consider a different name for this. It's not really "bytes available" if it's `storageInfo.bytesAvailable() * 0.95;`.

JamesMBallard (commented):
Coming along nicely. A few comments but nearly there.

Shared/configurations/ConfigurationController.cpp
dtedder: I think we run the risk of having Android OS shutting down the app by if we tried to read all bytes from the source with `QFile::readAll()`. I implemented a block size of 16KB (might be a bit small still...) and recalculate the the overall percentage downloaded so far. I'm using an integer type here as more or less a trick to limit the 'change' to only 1 whole percent at a time. More frequent updates than to the UI than that probably aren't necessary.

Shared/configurations/ConfigurationController.cpp
dtedder: Had this in there to ensure nothing caused integer division to happen. I think the ` * 100.0` at the end actually takes care of this already, yes.

Shared/configurations/ConfigurationController.cpp
JamesMBallard: Thanks, that makes more sense that we can crash the app on Android if we attempt to read GBs into memory at once.

JamesMBallard (approved):

Shared/configurations/ConfigurationController.cpp
dtedder: Actually this _was_ necessary to prevent integer division from happening when evaluating left to right. `bytesWritten` is integer type. I experimented with just using a float for bytesWritten instead here but the disassembly looked simpler with multiplying an integer type by 1.0 (at least on MacOS) to get the desired effect.

dtedder:
closing due to merge conflicts. reopened here #464

PR #461, opened 2025-08-12
Isolate setProperties to only those associated with a tool

dtedder:
- Add setProperty to AbstractTool
- Modify DsaController to call setProperty in onPropertyChanged signal
- Update tool controllers to implement setProperty and subsequently call their existing setProperties method if a relevant property was altered

@JamesMBallard

### DIFF
--- a/Shared/AbstractTool.h
+++ b/Shared/AbstractTool.h
@@ -43,8 +43,8 @@ public:
   virtual QString toolName() const = 0;
   virtual bool handleClick(const Esri::ArcGISRuntime::Point& pos);
 
-  virtual void setProperties(const QVariantMap& properties);
-  virtual void setProperty(const QString& propertyName, const QVariantMap& properties);
+  virtual void toolInitProperties(const QVariantMap& properties);
+  virtual bool shouldSetProperties(const QString& propertyName);
 
   virtual void setActive(bool active);
   bool isActive() const;
--- a/Shared/AbstractTool.h
+++ b/Shared/AbstractTool.h
@@ -17,9 +17,13 @@
 #ifndef ABSTRACT_TOOL_H
 #define ABSTRACT_TOOL_H
 
+// Qt headers
 #include <QObject>
 #include <QString>
 #include <QVariantMap>
+
+// STL headers
+#include <unordered_set>
 
 namespace Esri::ArcGISRuntime {
   class Error;
@@ -56,6 +60,7 @@ signals:
 
 protected:
   bool m_active = false;
+  static bool setContainsString(const std::unordered_set<QString>& strSet, const QString& str);
 };
 
 } // Toolkit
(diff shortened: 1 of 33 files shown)


Shared/AbstractTool.h
```diff
@@ -44,6 +44,7 @@ class AbstractTool : public QObject
   virtual bool handleClick(const Esri::ArcGISRuntime::Point& pos);
 
   virtual void setProperties(const QVariantMap& properties);
+  virtual void setProperty(const QString& propertyName, const QVariantMap& properties);
```
JamesMBallard: I like the new pattern. On the one hand, it allows each controller to opt-into which properties it cares about, and ignore the rest.

The problem I see now is that as a developer stepping into this codebase, I will not know why there are two of these or understand when I should call one over the other.

Here is a potential alternate approach. Let me know your thoughts. It may not work.

Make `setProperties` private in the class. Rename `setProperty` to this:
```c++
virtual bool shouldSetProperty(const QString& propertyName); // default behavior is always true
```

That way each controller can do some version of this:
```c++
bool AddLocalDataController::shouldSetProperty(const QString& propertyName)
{
  return propertyName == LOCAL_DATAPATHS_PROPERTYNAME;
}
```

And then the DsaController logic becomes:
```c++
if (tool->shouldSetProperty(propertyName))
{
  tool->setProperties(m_dsaSettings); // TBD ??
}
```

Conceptually the important parts are:
1. Each controller/tool can maintain its own list of properties it cares about. Your changes already implement that.
2. Each tool is only responsible to maintain the list, and the DsaContoller is still responsible for calling `setProperties`. `setProperties` becomes inaccessible from the subclasses, which further enforces this distinction.

Hopefully that would reduce ambiguity about which one to call, and when.

Shared/GridController.cpp
```diff
@@ -86,6 +86,13 @@ void GridController::setProperties(const QVariantMap& properties)
   setCoordinateFormat(properties["CoordinateFormat"].toString());
 }
 
+bool GridController::shouldSetProperties(const QString& propertyName)
+{
+  return (propertyName == PROPERTY_NAME_GRID_VISIBLE ||
+          propertyName == PROPERTY_NAME_GRID_COLOR_SCHEME ||
+          propertyName == QStringLiteral("CoordinateFormat"));
```
JamesMBallard: Some of these with longer lists of things to check may be good candidates for std::any_of

JamesMBallard (commented):
I like this better. One recommendation but it won't change things structurally.

Shared/GridController.cpp
dtedder: 1c0c22878b2e3fd317e62142fc44aaa6f7c67708

dtedder:
Merging to Grids branch. We can continue any further development in that PR.

Issue #458, opened 2025-07-26
Bug: Home button doesn't zoom to initial location in config file


kerryrobinson:
I noticed that when using other data configurations, hitting the Home button would zoom back to Monterey instead of the Initial Location set in the config file.  I can't recall if this was intentional originailly or just something we overlooked. 

Definition of Done: 
- [x] Hitting the Home icon on the map zooms to the location and view set by the Initial Location property in the DsaAppConfig.json file. 

Todo:
- [x] Implementation
- [x] Test
- [x] Merge

dtedder:
@kerryrobinson

Builds available here:
`\\apps-data\Data\apps\DSA\Builds\Issue-458`

kerryrobinson:
Working pretty well.  Only thing I noticed was that there is a 'Reset to Default Scene' button in the Scene Package chooser panel.  When I click that, I expect to be taken back to the original home in the config file, but it always goes to the default Monterey viewpoint.

dtedder:
Builds have been updated to include this fix: b30de3d67a5b9dfb2fc134a8b97628b1fbcb6852

kerryrobinson:
@dtedder I tested this update.  If I load a scene package, then "reset to default scene", I now notice that it takes me to the home location for the scene package, not the configuration's home/default location.  This is better than defaulting to Monterey, but I think it should go back to the configuration's home location. I remember now that we do overwrite some settings of the config file with the scene package info. 
We should discuss whether it's possible to save the original config's home location so that if the "reset to default scene" is clicked, all the original properties are written to the config file.  Or, writing out scene package info to a different config file so it doesn't overwrite and corrupt the one associated with the configuration.

kerryrobinson:
Can close this issue, but create a new one to address how to best support working with mobile scene packages and with DSA configurations.

PR #462, opened 2025-08-18
Read from the config file each call to 'home'

dtedder:
add a reference to the DsaController object on the ToolManager.
make the readInitialLocation method public but marked as const
use the result of the function for home instead of initialViewpoint on the scene which might not match what users have configured based on the startup sequence and packages loaded.

#458

### DIFF
--- a/Handheld/Handheld.cpp
+++ b/Handheld/Handheld.cpp
@@ -26,7 +26,6 @@
 
 // DSA headers
 #include "DsaController.h"
-#include "ToolManager.h"
 #include "ToolResourceProvider.h"
 
 using namespace Esri::ArcGISRuntime;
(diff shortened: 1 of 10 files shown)


Shared/DsaController.cpp
```diff
@@ -145,6 +145,7 @@ Scene* DsaController::scene() const
 void DsaController::init(GeoView* geoView)
 {
   ToolResourceProvider::instance()->setGeoView(geoView);
+  ToolManager::instance().setDsaController(this);
```
JamesMBallard: If there is `ToolManager::instance()`, then to me that says there is only one ToolManager since we're using the singleton pattern. Can we (or should we) make that same assumption for the `DsaController`? Should there be a static `DsaController::instance()` method? 

I do wonder if that if that would avoid the need to have objects pass themselves to other objects like this.

Shared/NavigationController.cpp
```diff
@@ -116,6 +117,17 @@ void NavigationController::zoomToInitialLocation()
   if (!m_is3d)
     return;
 
+  // read the users defined viewpoint from the config file
```
JamesMBallard, line 120: Is there a larger issue at play that we need to understand? I don't understand why calling to the `dsaController` solves this problem.

Shared/ToolManager.h
```diff
@@ -65,6 +69,7 @@ class ToolManager : public QObject
   ToolManager();
 
   ToolsList m_tools;
+  const DsaController* m_dsaController;
```
JamesMBallard:
```suggestion
  const DsaController* m_dsaController = nullptr;
```

Shared/DsaController.cpp
dtedder: Agreed. `DsaController::instance()` singleton would be a cleaner solution.

Shared/NavigationController.cpp
dtedder: All instances of the GeoView are set to the point for Monterey initially:
https://github.com/Esri/dynamic-situational-awareness-qt/blob/8bcc8f5aaacae69c7294563afa5b7e7769e44c03/Shared/DsaController.cpp#L102

Based on the docs, once the [initialviewpoint](
https://developers.arcgis.com/qt/cpp/api-reference/esri-arcgisruntime-geomodel.html#initialViewpoint) is set and a scene has been loaded, it cannot be modified. The code appears to contain some logic that checks during the loading process for the initial load itself. But for some reason the setInitialViewpoint is never actually called again, just read from and written back out to the config file as JSON. There are some comments that maybe imply because we are using MSPK, the setting of this value is not available to us in the code?
https://github.com/Esri/dynamic-situational-awareness-qt/blob/8bcc8f5aaacae69c7294563afa5b7e7769e44c03/Shared/DsaController.cpp#L593

JamesMBallard: That comment clarifies things. The `initialViewpoint` property is read and applied when a Scene is loaded. If you're using a mobile scene package, all the scenes accessed from `MobileScenePackage::scenes()` are already loaded, so there is no opportunity to change the `initialViewpoint` property since it's already too late at that point.

Handheld/Handheld.cpp
```diff
@@ -36,71 +35,52 @@ namespace Handheld {
 
 Handheld::Handheld(QQuickItem* parent /* = nullptr */):
   QQuickItem(parent),
-  m_controller(new DsaController(this))
+  m_controller(DsaController::instance())
```
JamesMBallard, line 38: Is `m_controller` even needed anymore?

Handheld/Handheld.h
```diff
@@ -45,8 +45,8 @@ class Handheld : public QQuickItem
   void errorOccurred(const QString& message, const QString& additionalMessage);
 
 private:
-  Esri::ArcGISRuntime::SceneQuickView*    m_sceneView = nullptr;
-  DsaController*                          m_controller = nullptr;
+  Esri::ArcGISRuntime::SceneQuickView* m_sceneView = nullptr;
+  DsaController& m_controller;
```
JamesMBallard, line 49: Hopefully this is moot, but it's dangerous to hold references like this.

Shared/DsaController.cpp
```diff
@@ -129,6 +128,13 @@ DsaController::~DsaController()
   saveSettings();
 }
 
+DsaController& DsaController::instance()
+{
+  static DsaController instance{};
+
+  return instance;
+}
```
JamesMBallard, line 136: I would return a pointer and use https://doc.qt.io/qt-6/qglobalstatic.html#Q_GLOBAL_STATIC for the lifetime management. It introduces new risks to have a local static like this, and `Q_GLOBAL_STATIC` can help mitigate those risks.

This is probably being overly cautious since other singletons don't do this, but all the more reason we may want to consider doing it now. We want to avoid https://en.cppreference.com/w/cpp/language/siof.html.

JamesMBallard (commented):
I recommend using `Q_GLOBAL_STATIC` for newly added singletons.

PR #463, opened 2025-08-20
add a private static const pointer to *this that can be accessed thro…

### LINKED ISSUE #458: Bug: Home button doesn't zoom to initial location in config file

dtedder:
Switch to static pointer to instance for DsaController

### DIFF
--- a/Shared/DsaController.cpp
+++ b/Shared/DsaController.cpp
@@ -69,8 +69,6 @@ QJsonObject defaultViewpoint();
 
 } // namespace
 
-const DsaController* DsaController::s_instance = nullptr;
-
 /*!
   \class Dsa::DsaController
   \inmodule Dsa
@@ -121,6 +119,7 @@ DsaController::DsaController(QObject* parent):
     firstLoad = false;
   });
 
+  Q_ASSERT(s_instance == nullptr); // there should never be more than one DsaController created
   s_instance = this;
 }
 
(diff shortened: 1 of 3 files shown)


Shared/DsaController.cpp
```diff
@@ -118,6 +120,8 @@ DsaController::DsaController(QObject* parent):
     updateInitialLocationOnSceneChange(firstLoad);
     firstLoad = false;
   });
+
+  s_instance = this;
```
JamesMBallard, line 123: As a safety check, you can assert if this is not null.

```c++
Q_ASSERT(s_instance == nullptr);
s_instance = this;
```

That would help us identify (in debug builds) if someone has mistakenly allocated more than one `DsaController`

Shared/DsaController.h
```diff
@@ -61,6 +64,7 @@ private slots:
   void errorOccurred(const QString& message, const QString& additionalMessage);
 
 private:
+  static const DsaController* s_instance;
```
JamesMBallard:
```suggestion
  static inline const DsaController* s_instance = nullptr;
```

'inline' in this context means single definition so you don't need do add any initialization code to the .cpp file for it.

dtedder:
#458

PR #464, opened 2025-08-21
Dtedder/402 download custom configurations fixed

### LINKED ISSUE #402: Add ability to download and manage custom DSA packages
As a follow-up to https://github.com/Esri/dynamic-situational-awareness-qt/issues/375: 

Requirements: 
- Add a "Custom Data" option to the new tool "Download Data", in addition to "DSA Demo Data"
- Allows the user to point to their own location where their data zip file is. this could be:
  - A portal item (Enterprise or Online - but the URL goes to an item endpoint that contains a zip file)
    - need to consider the auth challenge.  only support public items initially?
  - URL - a URL to the location of the zip file
  - File path - file path to the location of the zip file
- Data is unzipped into the DSA data folder, according to the platform it's deployed on
- If data already exists, ask the user if they want to wipe it first, otherwise cancel/abort!
- Allow the DSA user to switch between DSA packages
  - store the name of the folder in a json file at the root DSA folder so it knows which one to start up.
  - expose the DSA Package switcher in the DSA Packages panel

Definition of Done:   
- [x] "Custom Data" option is added to the new tool "Download Data", in addition to "DSA Demo Data"
- [x] Option to browse to A portal item (Enterprise or Online - but the URL goes to an item endpoint that contains a zip file)
- [x] Option to enter a URL - a URL to the location of the zip file
- [x] Option to browse to a File path - file path to the location of the zip file
- [x] Data is unzipped into the DSA data folder, according to the platform it's deployed on
- [x] If data already exists, ask the user if they want to wipe it first, otherwise cancel/abort!
- [x] User can switch between DSA packages
- [x] User can delete the configuration and data
  - [x] When hit delete the first time, it'll ask if you want to remove the files... if so, then it'll remove the files. If you click it again, it'll remove the configuration entry. 
- [x] If you try to download a file and the URL/link is invalid, you get an error message the option to delete the file.  
- [x] If you try to download a zip that does not include a DsaAppConfig.json file, you get an error message and the download process stops. 
- [ ] Validate it works with various valid/invalid URLs and content 

Known limitations:
- URLs to some cloud services might not work.  Permalinks from S3 will work (they are temporary and include tokens,etc).  An alternative is to download the file to the device first and then load. 

Todo:
- [x] Design discussed and agreed
- [x] Implementation done
- [x] Tested 
- [x] Merged and verified 
- [x] Doc updated (include the format/scheme we expect the data to be in the zip file)
  - [x] Steps detailing the zip folder structure expected by the tool
  - [x] Requirement of the DsaAppConfig.json file and its expected location
  - [x] Instructions for using 'local' files on mobile from the Downloads or Documents folders

dtedder:
@JamesMBallard 

Issue #402
Continuation of PR #460
Fixes merge conflicts from #459 and adds new detail from discussion with @kerryrobinson during this weeks catchup.


JamesMBallard (approved):

dtedder:
@JamesMBallard this one is ready for re-review again. Starting at this commit: https://github.com/Esri/dynamic-situational-awareness-qt/pull/464/commits/c3e4a634edf0ef7e0423a89011e8b04989244cd2.

Some cleanup of the gradle build files, updating some logic for accessing the app directory on iOS so users can copy config files from their downloaded files, and some updates to the logic for when any failures happen either in download or unzipping.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,77 +79,347 @@ void ConfigurationController::select(int index)
...
+      portalItem_doneLoading(portal, portalItem, configurationName, error);
+    });
+    portalItem->load();
+    return;
   }
+
+  // make a head request to check the availabiity of the network as well as get the final size of the package to be downloaded
+  QNetworkRequest zipRequest{configurationUrl};
+  auto* headReply = m_networkAccessManager.head(zipRequest);
+  headReply->ignoreSslErrors();
```
JamesMBallard: Should we make this determination for users? It could be a security risk. I would reconsider unless this solves a specific problem you are hitting.

Shared/configurations/ConfigurationListModel.cpp
```diff
@@ -46,11 +46,30 @@ void ConfigurationListModel::cancel(int index)
...
 
+Configuration ConfigurationListModel::at(int index) const
+{
+  if (index < 0 || index >= m_configurations.size())
+    return Configuration{};
+
+  return m_configurations[index];
+}
+
+Configuration ConfigurationListModel::byName(const QString& name) const
```
JamesMBallard, line 62: These are all copied value objects. I want to double check my understanding. A potential pitfall is if calling code is expecting any kind of pointer/reference semantics to work.

For example, this code does nothing (assignment to a temporary). Is that correct?
```c++
model->at(0).setName("Some new name");
```

Shared/utilities/ZipHelper.cpp
```diff
@@ -158,6 +158,13 @@ bool ZipHelper::extractCurrentFile(const QString& outputFilePath)
 
             emit extractProgress(fileName, outputFilePath, percent);
 
+            // check for a new percentage but only by whole numbers
+            auto oldPercentTotal = (int)percentTotal();
+            m_bytesExtracted += result;
+            auto newPercentTotal = (int)percentTotal();
```
JamesMBallard: Avoid c style casts when possible. Prefer static_cast

Shared/utilities/ZipHelper.h
```diff
@@ -106,12 +107,15 @@ class ZipHelper : public QObject
   void zwClose(const QString& globalComment = QString());
   bool exists() const;
   bool zrOpen();
+  qreal percentTotal();
```
JamesMBallard: Can this be const?

Shared/qml/Options.qml
```diff
@@ -321,7 +335,7 @@ Rectangle {
                                 verticalCenter: parent.verticalCenter
                             }
                             checked: model.Selected
-                            enabled: model.Downloaded
+                            enabled: model.Downloaded && model.Extracted
```
JamesMBallard, line 338: Nitpick, but does `model.Extracted` imply `model.Downloaded`?

Shared/qml/Options.qml
```diff
@@ -20,11 +20,25 @@ import QtQuick.Controls.Material
...
 
 Rectangle {
     id: optionsRoot
     property real scaleFactor: (Screen.logicalPixelDensity * 25.4) / (Qt.platform.os === "windows" || Qt.platform.os === "linux" ? 96 : 72)
 
+    Connections {
+        target: configurationController
+        function onConfigurationDownloadFailed(configurationIndex, configurationName, message) {
+            if (configurationIndex === -1)
+                return;
```
JamesMBallard, line 33: When could this happen? Is -1 indicative of a special failure?

Shared/qml/AddConfigurationTool.qml
```diff
@@ -0,0 +1,170 @@
...
+        ComboBox {
+            id: comboResourceType
+            Layout.fillWidth: true
+            Layout.leftMargin: 5
+            Layout.topMargin: 5
+            Layout.rightMargin: 5
+            model: configTypeNames
+            onCurrentValueChanged: {
+                txtUrl.validator = configRegex[configTypeNames.indexOf(comboResourceType.currentValue)]
+                txtUrl.placeholderText = configHints[configTypeNames.indexOf(comboResourceType.currentValue)]
```
JamesMBallard, line 74: Very nice 👍

Shared/qml/AddConfigurationTool.qml
```diff
@@ -0,0 +1,170 @@
...
+
+DsaPanel {
+    id: addConfigurationRoot
+    width: 272 * scaleFactor
+    title: qsTr("Add Configuration")
+
+    property RegularExpressionValidator configRegexName: RegularExpressionValidator { regularExpression: /^[A-Za-z0-9_]+$/ }
+    property RegularExpressionValidator configRegexItem: RegularExpressionValidator { regularExpression: /^https:\/\/.+\/item\.html\?id=[A-Fa-f0-9]{32}?$/ }
+    property RegularExpressionValidator configRegexWeb: RegularExpressionValidator { regularExpression: /^https:\/\/.+/ }
+    property RegularExpressionValidator configRegexFile: RegularExpressionValidator { regularExpression: /^file:\/\/.+/ }
```
JamesMBallard, line 33: Can any of these be readonly properties?

Shared/configurations/ConfigurationController.cpp
dtedder: I had this in for testing purposes but left it in probably assuming too much. Our typical user _might_ face the same situation as I did where they are just downloading from a local server without proper SSL... But even if that's the case, the scenario of downloading to the device and moving/copying over to the app folder is probably a better and even simpler option. I will remove.

Shared/configurations/ConfigurationListModel.cpp
dtedder: Yes, these were intended to be copies for the convenience of getting all the information in one call (name, QUrl, QString url). The collection can change since configurations can be removed/added while other's might still be downloading/extracting. Modifications must be made using the setData or setDataByName methods on the list model.

Shared/qml/Options.qml
dtedder: If the indexByName function fails to find a configuration with the name provided, it will return -1. I don't have a concrete example or test case for this. But with the async behaviors that have been introduced in this PR, I can imagine there is something that hasn't been tried yet that could produce a call to this method with a name of a configuration that has already been removed.

Shared/qml/Options.qml
dtedder: I think you are correct, yes. I do not think there is any state in the application UI right now that needs to be expressed for a case where a configuration may have been downloaded but not extracted or even vice versa (we actually delete the download once it's extracted for space). These two operations should really be combined into one overall atomic state of available/not-available external to the class. This instance can probably just be switched to just `model.Extracted`. But I'm hesitant to make that change right now since it's been working as expected in mine and @kerryrobinson 's testing so far. I would prefer a follow up issue/ticket to do some refactoring/cleanup on items that evolved overtime in the code since the start of the development on the feature.

Shared/qml/Options.qml
JamesMBallard: My naive assumption is that the event is not fired if the index is invalid, so that should not be a concern of the called code, but the calling code. You are more familiar with the implementation details, so I will defer to you. It's not harmful to keep defensive code.

Shared/qml/Options.qml
JamesMBallard: Not necessary to make any changes. I was just thinking out loud mostly.

Shared/configurations/ConfigurationListModel.cpp
```diff
@@ -196,12 +186,20 @@ bool ConfigurationListModel::add(const QString& name, const QString& url, bool s
...
+  for (qsizetype index = 0; index < m_configurations.count(); ++index)
+  {
+    if (m_configurations[index].name() == configurationName)
+    {
+      beginRemoveRows(QModelIndex{}, index, index);
+      m_configurations.removeAt(index);
+      endRemoveRows();
+      return true;
+    }
+  }
```
JamesMBallard, line 200: You could consider std::find_if here to avoid the loop, but optional.

JamesMBallard (approved):
Another recommendation, but optional.

Shared/configurations/ConfigurationListModel.cpp
dtedder: The way I have it written here, I think I require the integer for the index. But maybe there's some property on QList I'm overlooking? I should probably re-write this entirely using std::vector<Configuration>... But that ended up making too many changes across all the other methods that use this QList. It might be good to just have a ticket for refactoring through out places in the app wherever appropriate to use STL containers.

JamesMBallard: You are right, `std::find_if` won't get you the index and you need the index. Disregard my comment.

JamesMBallard: >It might be good to just have a ticket for refactoring through out places in the app wherever appropriate to use STL containers.

Guillaume on the Qt team did some research on this a few years ago to see if there were enough benefits (or any) to switch to STL containers. The short version was no, not enough benefits. The Qt container classes are good, and aside from some detach concerns, there wasn't enough benefit to transition away from them.

PR #465, opened 2025-09-04
add pdf as extension for rasters

### LINKED ISSUE #418: Packages tool (Handheld): Overlays list does not show list of layers from the MSPK

### LINKED ISSUE #456: Geospatial PDF Support
The Qt SDK 200.8 supports PDF files as raster layers.  We should make sure to include this as another valid raster layer file format for Operational Layers. 

Dependent on #457 

Sample data: https://devtopia.esri.com/runtime/milano/blob/main/References/GeoPDF/GeoPDF_list.md
PDFs also can be found here: \\apps-data\data\apps\DSA\SampleData\DSA\MontereyFull\OperationalData:
- `adobe_style_geospatial.pdf`: this is a valid georeferenced PDF, it just only shows 2 map frames and grids, and is located in Nova Scotia. 
- `kilauea-volcano.pdf`: does not have an SR that we can interpret
- `Ops_Planning_land_20160930_Soberanes_CA....pdf`: (valid) large Cal Fire map from the Soberanes fire in 2016 near Monterey
- `pio_arch_e_land_0825_2045_River....pdf`: (valid) large Cal Fire map from the River fire in 2020
- `UnknownSR.pdf`: actually displays somewhere in the Pacific

Definition of Done:
- [x] User can browse to a .pdf file on disk when adding an Operational Layer (add .pdf as valid file extension)
- [x] PDF added as a raster layer
- [x] When using the Add tool to add a raster file that does not have a spatial reference, Show error message to user and do not add to the Layer List
- [x] If the config json file is modified to point to a layer that is invalid, the layer will show in the list, but the layer name will be in red, signalling to the user there is an issue and it won't display. 

Todo:
- [x] Implementation
- [ ] Test

dtedder:


Shared/AddLocalDataController.cpp
```diff
@@ -718,23 +719,44 @@ void AddLocalDataController::createFeatureLayerShapefile(const QString& path, in
...
+        rasterLayer->deleteLater();
+        raster->deleteLater();
+        emit toolErrorOccurred(QString("The raster (%1), did not contain spatial reference information.").arg(path), failedAdditionalMsg);
+        return;
+      }
+
+      if (auto* operationalLayers = ToolResourceProvider::instance()->operationalLayers(); operationalLayers)
+        operationalLayers->append(rasterLayer);
+
+    }, Qt::SingleShotConnection);
```
JamesMBallard, line 755: Nice usage of Qt::SingleShotConnection 👍.

Shared/LayerCacheManager.cpp
```diff
@@ -442,7 +445,43 @@ void LayerCacheManager::connectSignals()
...
+              operationalLayers->removeOne(rasterLayer);
+              rasterLayer->raster()->deleteLater();
+              rasterLayer->deleteLater();
+              emit toolErrorOccurred(QString("The raster (%1), failed to load. [%2][%3]").arg(rasterPath, loadError.message(), loadError.additionalMessage()), rasterFailedAdditionalMsg);
+              return;
+            }
+
+            // check the spatial reference for the new layer for:
+            // - non-georeferenced PDFs may have been added
+            if (const auto spatialReference = rasterLayer->spatialReference(); spatialReference.isEmpty())
```
JamesMBallard: You don't need to hold onto the object at all here. 

```suggestion
            if (rasterLayer->spatialReference().isEmpty())
```

Shared/LayerCacheManager.cpp
```diff
@@ -442,7 +445,43 @@ void LayerCacheManager::connectSignals()
...
+            {
+              operationalLayers->removeOne(rasterLayer);
+              rasterLayer->raster()->deleteLater();
+              rasterLayer->deleteLater();
```
JamesMBallard: Can you make one the parent of the other? That would avoid the multiple deleteLater calls on each one, but it may be more trouble than it's worth compared to being explicit like this.

Take this with a grain of salt. The suggestion may be worse.

JamesMBallard (commented):
Looks pretty good. A few comments but not much to complain about here.

Shared/LayerCacheManager.cpp
dtedder: Good catch. Yep, not using this anywhere else below. Can't remember why I was storing a copy of this to begin with so I'll switch it out.

Shared/LayerCacheManager.cpp
dtedder: This turned out to actually be a requirement. Removing layers is possible from the `[...]` menu in the ToC. Without setting the parent here was leading to memory growth that never came down. Have a look at this commit 743dce6c28e44dcd7bd54a03f8b4085250632dde. Adding and removing the same .dted file (one of our larger test rasters) over and over made it pretty clear. It may not have been noticeable before.

Shared/qml/TableOfContents.qml
```diff
@@ -62,17 +62,21 @@ DsaPanel {
...
             imageUrl: imageSourceForGeomType(index)
             imageVisible: true
             imageFrameVisible: false
             menuIconVisible: true
-            mainText: name && name !== "" ?
-                          name :
-                          toolController.alternateName(index)
-            onItemCheckedChanged: layerVisible = itemChecked
+            property bool spatialReferenceOk: toolController.spatialReferenceOk(index)
+            mainText: name && name !== "" ? name : toolController.alternateName(index)
```
JamesMBallard: presumably 'name' is a string, so it can't be null. It can be undefined but that seems uncommon to check for with strings. Do you think `name &&` is needed?

Shared/AddLocalDataController.cpp
```diff
@@ -718,23 +719,46 @@ void AddLocalDataController::createFeatureLayerShapefile(const QString& path, in
...
+      if (!loadError.isEmpty())
+      {
+        rasterLayer->deleteLater();
+        emit toolErrorOccurred(QString("The raster (%1), failed to load. [%2][%3]").arg(path, loadError.message(), loadError.additionalMessage()), failedAdditionalMsg);
+        return;
+      }
+
+      // check the spatial reference for the new layer for:
+      // - non-georeferenced PDFs may have been added
+      if (const auto spatialReference = rasterLayer->spatialReference(); spatialReference.isEmpty())
```
JamesMBallard:
```suggestion
      if (rasterLayer->spatialReference().isEmpty())
```

JamesMBallard (approved):
Two more things, both minor. No need for another review but I am happy to review again if you want to change anything.

Shared/qml/TableOfContents.qml
dtedder: Didn't notice this when I adjusted the line. I'll remove the unnecessary null check.

dtedder:
Thanks @JamesMBallard. I made another commit 65dd61b5a313794ceae2acb77c93920d47674728 switching to `deleteLater` instead of using `std::unique_ptr` based on your comment here: https://github.com/Esri/dynamic-situational-awareness-qt/pull/472#discussion_r2446019208.

May not be any harm. But as you suggest, probably better not to mix the two schemes. The layers are QObjects with a parent set. `deleteLater` seems more appropriate here too.

JamesMBallard (approved):

PR #471, opened 2025-10-07
Aji13187/7587 deprecation warnings

aji13187:
Suppressed deprecation warnings in Handheld and Vehicle.

image


bri12415:
You may want to run this by @dtedder, I know this fix was suggested before but they wanted to be able to preserve some of the warnings for future work. This fix would disable all deprecation warnings.

aji13187:
@dtedder - 

> You may want to run this by @dtedder, I know this fix was suggested before but they wanted to be able to preserve some of the warnings for future work. This fix would disable all deprecation warnings.

Could you clarify this? What deprecation need to be preserved? 
This change won't suppress Qt deprecation warnings, this suppresses only ArcGIS SDK warnings.

dtedder:
@aji13187 
This appears to be directly related to using precompiled header from qmake. If I remove the PCH defines from the .pro files (or target Android which doesn't use them), I get only the errors that are related to the DSA project files:
image

Using `pch.hpp` gives warning from the toolkit files also:
image

The [pch file](https://github.com/Esri/dynamic-situational-awareness-qt/blob/main/Shared/pch.hpp) includes the following. So far as I can tell, there's nothing directly included here that references anything related to the warnings for files in the toolkit.
```
#ifndef DSA_PCH_HPP
#define DSA_PCH_HPP

#ifdef __cplusplus

#include "AlertCondition.h"
#include "AlertFilter.h"
#include "AlertLevel.h"
#include "AlertTarget.h"

#include "AbstractTool.h"
#include "ToolManager.h"
#include "ToolResourceProvider.h"

#include "GeoElement.h"
#include "GeometryTypes.h"
#include "Point.h"
#include "Scene.h"
#include "SceneView.h"

#include <QAbstractListModel>
#include <QHash>
#include <QList>
#include <QObject>
#include <QPointer>
#include <QVariantMap>

#endif // __cplusplus

#endif // DSA_PCH_HPP
```

dtedder:
@aji13187
An example would be the warnings for the `geometryChanged()` signal in `GeoElementUtils.cpp`. Those are valid and we want to see them until we address them in the project code.

I have not been able to determine why these warnings only show up if the build is using precompiled headers. But for now I think we can just ignore the unrelated warnings.

aji13187:
> why these warnings only show up if the build is using precompiled headers?

The pch file indirectly includes `AuthenticationManager` transitively from other includes in the pch.hpp.
When building with precompiled header enabled, the [Deprecated.h](https://devtopia.esri.com/runtime/qt/blob/main/api/qt_cpp/Include/Deprecated.h) file is parsed in the very beginning before `QRT_DISABLE_DEPRECATED_WARNINGS` is defined in toolkit cpp files. So `[[deprecated]]` is enabled for the build.
This is causing the toolkit warnings.

When building without PCH, the order of includes and macro defines are synchronised and working properly.

I don't have a  good solution yet. @GuillaumeBelzPro, @bri12415 any suggestions?

aji13187:
One solution is to remove the API and toolkit headers from `pch.hpp`, So that the toolkit warnings wont appear in DSA build.  PR is updated for that.

An another solution is to use below pragma in the .cpp files around the deprecated code in toolkit/API.
```
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
#elif defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
#endif

// code

#if defined(__clang__) || defined(__GNUC__)
#pragma clang diagnostic pop   // (or GCC pop)
#endif
```

Shared/pch.hpp
```diff
@@ -25,21 +41,10 @@
 #include "AlertLevel.h"
 #include "AlertTarget.h"
 
-// toolkit headers
-#include "AbstractTool.h"
-#include "ToolManager.h"
-#include "ToolResourceProvider.h"
-
-// C++ API headers
```
bri12415, line 33: I thought the DSA folks wanted to keep the API warnings but suppress the toolkit warnings. If that's the case, I don't think we need to remove the API headers from the pch, just the toolkit files. Is that not right? 

https://github.com/Esri/dynamic-situational-awareness-qt/pull/471#issuecomment-3386982505

bri12415 (approved):

dtedder:
@aji13187 ok, I see the effects now. Just for some clarification, there are comments here where the headers are include that refer to 'toolkit'. However, this is actually not correct. Several of the comments near these includes are not referencing anything to do with the official Esri toolkit. I've been removing them as I've encountered them.  The includes ultimately just break down into the following categories and we've been using this order since I've started working on DSA. Here's a sample below. I have not been _strictly_ adhering to the rule of including anything that's used. But in general I've been trying to do this wherever it makes sense. [cpp guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf10-avoid-dependencies-on-implicitly-included-names). I will go through and make some notes in the PR. Not all the includes need to be removed from the pch. Some can stay without causing the warnings. 

```
// PCH header (if in a cpp file)
#include "pch.hpp"

#include "CurrentSource.h" // the header for the current file

// C++ API headers (maps sdk headers)
#include "Camera.h"
#include "Field.h"
...
#include "Graphic.h"

// Qt headers
#include <QFile>
...
#include <QString>

// DSA headers (application files)
#include "MarkupConstants.h"
#include "ToolResourceProvider.h"

// STL headers
#include <vector>
```

Shared/pch.hpp
```diff
@@ -1,3 +1,19 @@
+// COPYRIGHT 2025 ESRI
```
dtedder: Let's remove added file headers for now. I don't think this is the header we want for DSA. But if we do want to introduce a new header, let's do it in a separate PR.

Shared/pch.hpp
```diff
@@ -25,21 +41,10 @@
...
 #include "AlertTarget.h"
 
-// toolkit headers
-#include "AbstractTool.h"
-#include "ToolManager.h"
-#include "ToolResourceProvider.h"
-
-// C++ API headers
-#include "GeoElement.h"
-#include "GeometryTypes.h"
```
dtedder, line 35: These two can stay

Shared/pch.hpp
```diff
@@ -25,21 +41,10 @@
 #include "AlertLevel.h"
 #include "AlertTarget.h"
 
-// toolkit headers
-#include "AbstractTool.h"
-#include "ToolManager.h"
```
dtedder, line 30: These can stay.

Vehicle/Vehicle.pro
```diff
@@ -22,6 +22,7 @@ CONFIG += c++17
 
 ARCGIS_RUNTIME_VERSION = 200.8.0
 DEFINES += ARCGIS_MAPS_SDK_VERSION=$$ARCGIS_RUNTIME_VERSION
+
```
dtedder: if this is a formatting rule, let's make it in a dedicated PR for formatting cleanup

Shared/GridController.cpp
```diff
@@ -28,6 +28,8 @@
 #include "MapViewTypes.h"
 #include "MGRSGrid.h"
 #include "SimpleLineSymbol.h"
+#include "ToolManager.h"
```
dtedder:
```suggestion
```
Not needed. This can still come from pch.

Shared/alerts/AlertConditionData.cpp
```diff
@@ -49,7 +50,7 @@ namespace Dsa {
   \sa AlertSource
   \sa AlertTarget
   \sa AlertCondition
-  */
+ */
```
dtedder: formatting. save for another PR

Vehicle/Vehicle.pro
aji13187: These are caused by the git hooks rules, automatically updated during the commits.
I'll try to bypass them.

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -41,6 +41,8 @@
 #include "LayerListModel.h"
 #include "MapTypes.h"
 #include "QueryParameters.h"
+#include "Scene.h"
+#include "SceneView.h"
```
dtedder, line 45: These includes are good. But remove the formatting on comments in the rest of the file.

Shared/alerts/GraphicAlertSource.cpp
```diff
@@ -1,3 +1,21 @@
+// COPYRIGHT 2025 ESRI
```
dtedder: same as others.

Shared/analysis/CombinedAnalysisListModel.cpp
```diff
@@ -20,6 +20,7 @@
 #include "CombinedAnalysisListModel.h"
 
 // dsa app headers
+#include "GeoElement.h"
```
dtedder:
```suggestion
```
ok to get from pch

Shared/analysis/CombinedAnalysisListModel.cpp
```diff
@@ -49,23 +51,23 @@ namespace Dsa {
   The model returns data for the following roles:
   \table
     \header
-        \li Role
-        \li Type
-        \li Description
+      \li Role
```
dtedder: same as other formatting comments

Shared/analysis/GeoElementViewshed360.cpp
```diff
@@ -25,6 +25,7 @@
 // C++ API headers
 #include "AnalysisOverlay.h"
 #include "AttributeListModel.h"
+#include "GeoElement.h"
```
dtedder:
```suggestion
```

Shared/configurations/ConfigurationController.cpp
```diff
@@ -38,6 +38,7 @@
 #include "NetworkRequestProgress.h"
 #include "Portal.h"
 #include "PortalItem.h"
+#include "ToolManager.h"
```
dtedder:
```suggestion
```

Shared/markup/MarkupBroadcast.cpp
```diff
@@ -23,6 +23,7 @@
 // dsa app headers
 #include "DataListener.h"
 #include "DataSender.h"
+#include "ToolManager.h"
```
dtedder:
```suggestion
```

Shared/markup/MarkupLayer.cpp
```diff
@@ -30,10 +30,13 @@
 #include "Graphic.h"
 #include "GraphicListModel.h"
 #include "GraphicsOverlay.h"
+#include "Point.h"
```
dtedder:
```suggestion
```
Cant find Point.h used here.

Shared/messages/MessageFeed.cpp
```diff
@@ -29,6 +29,7 @@
 #include "DynamicEntityObservationInfo.h"
 #include "Error.h"
 #include "Field.h"
+#include "GeometryTypes.h"
```
dtedder:
```suggestion
```
in pch

Shared/packages/OpenMobileScenePackageController.cpp
```diff
@@ -25,6 +25,7 @@
 #include "Item.h"
 #include "MapTypes.h"
 #include "MobileScenePackage.h"
+#include "ToolResourceProvider.h"
```
dtedder: You can actually swap line 40 with this one. Since the ToolManager.h is still ok to be in the pch, it's not needed anymore. One less line added to the file.

dtedder (changes requested):
Few comments. But looks like the overall goal can be accomplished by keeping some of the Esri headers out of the PCH. Thanks for getting to the root cause!

dtedder (approved):

Shared/pch.hpp
aji13187: It we cannot have any header which directly or iteratively includes [Deprecated.h](https://devtopia.esri.com/runtime/qt/blob/main/api/qt_cpp/Include/Deprecated.h). It can be API or toolkit headers.

aji13187:
@dtedder - Do we need to cherry pick this patch to main?

dtedder:
> @dtedder - Do we need to cherry pick this patch to main?

Not necessary right now, no.

PR #472, opened 2025-10-16
Dtedder/429 replace popup manager

### LINKED ISSUE #429: Remove PopupManager [deprecated]
DSA makes use of the PopupManager in the IdenfityController tool. This has been deprecated at 200.7 and will need to be adjusted to use the PopupElement design pattern.

> Note: we still get some warnings for PopupManager, doesn't look like there is a way to avoid (until it's actually removed from the toolkit)

Acceptance Criteria:
- [x] popups working as expected (although they look a little different) on feeds and layers
- [x] existing tests in the test plan should still pass
- [x] if popups are configured on the layer (mspk, slpk), it should display as authored (need test data)

dtedder:
Removed the PopupManager, PopupStackView and AuthenticationManager.

Replaced with custom QML using PopupView and Authenticator

### DIFF
--- a/Handheld/main.cpp
+++ b/Handheld/main.cpp
@@ -19,7 +19,6 @@
 #include "pch.hpp"
 
 // C++ API headers
-#include "PopupManager.h"
 #include "SceneQuickView.h"
 
 // Qt headers
@@ -109,7 +108,6 @@ int main(int argc, char *argv[])
 
   // Register the map view for QML
   qmlRegisterType<SceneQuickView>("Esri.ArcGISRuntime.OpenSourceApps.Handheld", 1, 1, "SceneView");
-  qRegisterMetaType<PopupManager*>("PopupManager*");
 
   // Register the Handheld (QQuickItem) for QML
   qmlRegisterType<Dsa::Handheld::Handheld>("Esri.ArcGISRuntime.OpenSourceApps.Handheld", 1, 1, "Handheld");
--- a/Handheld/main.cpp
+++ b/Handheld/main.cpp
@@ -19,6 +19,7 @@
 #include "pch.hpp"
 
 // C++ API headers
+#include "ArcGISRuntimeEnvironment.h"
 #include "SceneQuickView.h"
 
 // Qt headers
@@ -144,6 +145,7 @@ int main(int argc, char *argv[])
   view.engine()->addImageProvider(QStringLiteral("packages"), new Dsa::PackageImageProvider);
 
   Esri::ArcGISRuntime::Toolkit::registerComponents(*view.engine());
+  Esri::ArcGISRuntime::ArcGISRuntimeEnvironment::setUseLegacyAuthentication(false);
 
 #ifndef DEPLOYMENT_BUILD
   // Add the import Path
--- a/Shared/IdentifyController.cpp
+++ b/Shared/IdentifyController.cpp
@@ -21,241 +21,95 @@
 
 // C++ API headers
 #include "AttributeListModel.h"
+#include "FieldsPopupElement.h"
 #include "GeoElement.h"
 #include "GeoView.h"
-#include "Graphic.h"
-#include "GraphicsOverlay.h"
-#include "IdentifyGraphicsOverlayResult.h"
-#include "IdentifyLayerResult.h"
-#include "LayerContent.h"
 #include "Popup.h"
-#include "PopupAttributeListModel.h"
 #include "PopupDefinition.h"
+#include "PopupElement.h"
 #include "PopupField.h"
 #include "PopupFieldFormat.h"
-#include "PopupManager.h"
+#include "PopupFieldListModel.h"
+#include "PopupTypes.h"
 
 // DSA headers
-#include "GraphicsOverlaysResultsManager.h"
-#include "LayerResultsManager.h"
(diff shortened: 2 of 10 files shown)


Handheld/main.cpp
```diff
@@ -146,6 +145,7 @@ int main(int argc, char *argv[])
   view.engine()->addImageProvider(QStringLiteral("packages"), new Dsa::PackageImageProvider);
 
   Esri::ArcGISRuntime::Toolkit::registerComponents(*view.engine());
+  Esri::ArcGISRuntime::ArcGISRuntimeEnvironment::setUseLegacyAuthentication(false);
```
JamesMBallard, line 148: Just to confirm, but DSA v.next is using 200.8.x, correct? This API is gone at 300.0.

Shared/IdentifyController.cpp
```diff
@@ -265,36 +118,50 @@ bool IdentifyController::addGeoElementPopup(GeoElement* geoElement, const QStrin
     return false;
 
   // create a new Popup from the geoElement
-  Popup* newPopup = new Popup(geoElement, this);
+  std::unique_ptr<Popup> newPopup{new Popup(geoElement, this)};
```
JamesMBallard: Not recommended to mix ownership between QObject parents and unique_ptr. Typically one wins (unique_ptr goes first), and then QObject is resilient against double deletes, but we should be explicit with ownership intent.


```suggestion
  std::unique_ptr<Popup> newPopup = std::make_unique<Popup>(geoElement);
```

Shared/IdentifyController.cpp
```diff
@@ -21,241 +21,94 @@
...
-void IdentifyController::onMouseClicked(QMouseEvent& event)
+bool isObjectIdFieldName(QStringView fieldName)
 {
-  // ignore the event if the tool is not active.
-  if (!isActive())
-    return;
-
-  // only consider left clicks (or taps) onn the view.
-  if (event.button() != Qt::MouseButton::LeftButton)
-    return;
```
JamesMBallard, line 175: Do we need to retain any of this logic?

Shared/IdentifyController.h
```diff
@@ -48,27 +57,22 @@ class IdentifyController : public AbstractTool
...
 private:
   bool addGeoElementPopup(Esri::ArcGISRuntime::GeoElement* geoElement, const QString& popupTitle);
+  Esri::ArcGISRuntime::Popup* popup() const;
+  bool canNext() const;
+  bool canPrev() const;
 
-  bool m_isBusy = false;
   double m_tolerance = 5.0;
-  QList<Esri::ArcGISRuntime::PopupManager*> m_popupManagers;
+  int m_currentPopup = -1;
```
JamesMBallard: I would go with `m_currentPopupIdx` or `m_currentPopupIndex`. I got confused looking at comparisons against ints elsewhere.

Shared/IdentifyController.cpp
```diff
@@ -21,241 +21,94 @@
...
-  \brief Show popups for all of the \a geoElementsByTitle.
-
-  A popup will be created for each \l Esri::ArcGISRuntime::GeoElement in the QHash,
-  with the string key as the title.
- */
-void IdentifyController::showPopups(const QHash<QString, QList<GeoElement*>>& geoElementsByTitle)
+void IdentifyController::prevPopup()
 {
-  if (geoElementsByTitle.isEmpty())
+  if (m_currentPopup < 1)
```
bri12415: Can you instead use your `canPrev()` and `canNext()` functions as safety checks for these `prevPopup()` and `nextPopup()` functions?

bri12415 (approved):

Handheld/main.cpp
dtedder: Yes. Still 200.8.

Shared/IdentifyController.h
dtedder: Yep, good call. I'll rename to make it clear.

Shared/IdentifyController.cpp
dtedder: At some point, the IdentifyController was [hidden](https://github.com/Esri/dynamic-situational-awareness-qt/commit/7f1089a53d3c6429371753e63ea657a38215cd72#diff-2a2f349f77519fceb17ab36c643df5415f0161baffba2fc5ce604f79288156fdR156) on the main toolbar. The functionality appears to have been migrated over into the ContextMenuController, which presents the menu depending on the click results. It presents Identify as an option if any GeoElements are in the results. I removed this code to reduce confusion. It is not used in Vehicle or Handheld and was ultimately duplicate of what's in ContextMenuController: https://github.com/Esri/dynamic-situational-awareness-qt/blob/b22f92be4dc4916f2c6f1d95d64fe7cda275861a/Shared/ContextMenuController.cpp#L130

Shared/IdentifyController.cpp
dtedder: Understood. Will switch to std::make_unique.

Shared/IdentifyController.cpp
dtedder: Yep. Good catch. They were written for the QML but they make things easier to read here as well.

JamesMBallard (approved):

PR #475, opened 2025-10-30
Dtedder/445 use relative paths in config

### LINKED ISSUE #445: Look into writing out relative paths in the DsaAppConfig file
As a follow-on from https://github.com/Esri/dynamic-situational-awareness-qt/issues/364, we need to determine if the app should store relative paths for layers when it writes them out to the DsaAppConfig file (relative to the config file itself). 

Pros:
- This will allow the app to run better directly from a USB key, so that you could add operational layers to the app, close it (persist the config file) then open it on a different machine or different USB slot that has a different drive letter.
- This makes it much easier to move and share configurations without the need to update the file paths every time

Cons: 
- Users might not be able to use operational data from a folder that is not in the same folder/location as the config file (would need to test).

Todo:
- [x] Test how relative paths work when the DSA app is using the DsaAppConfig file in `user`/ArcGIS/data/Runtime/DSA.    We know it works as expected when the data is in the same folder as the executable. 
- [x] Test accessing operational data from a folder that is not in the same location as the DsaAppConfig file
- [x] Decide whether to always persist relative paths for operational data (decision: no, can use both)
- [x] Implement (if necessary)
- [ ] update tests
- [ ] update doc

Story:
DSA will look for the DsaAppConfig.json file in two places:
- First, in the same folder as the exe
- Then, in the default user folder (ArcGIS/runtime/data, or the mobile device's sandbox location

DSA can now use relative paths (as well as full paths) to reference data in the DsaAppConfig.json (relative to the config file). The app won't change the format when persisting app settings to the file.  Relative paths are useful for "packaging" the config data with the app; full paths are useful when the config data is in a different location than the app. 

Acceptance criteria:
- [ ] DSA can read relative paths from DsaAppConfig.json , and will write relative paths to DsaAppConfig.json
- [ ] If config file has full paths, DSA will write out full paths (doesn't change from full to relative)

dtedder:
Key to using relative paths is just to set the application's current directory `Shared/DsaController.cpp:103`. Once this is done, the Maps SDK data sources can work with relative or absolute paths in the config file.

#445

### DIFF
--- a/Shared/AddLocalDataController.cpp
+++ b/Shared/AddLocalDataController.cpp
@@ -99,9 +99,6 @@ AddLocalDataController::AddLocalDataController(QObject* parent /* = nullptr */):
   AbstractTool(parent),
   m_localDataModel(new DataItemListModel(this))
 {
-  // add the base path to the string list
-  addPathToDirectoryList(".");
-
   // create file filter list
   m_fileFilterList = QStringList{allData(), rasterData(), geodatabaseData(),
       sceneLayerData(), tilePackageData(), shapefileData(), geopackageData(),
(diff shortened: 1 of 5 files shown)


ldanzinger (approved):
lgtm

JamesMBallard (approved):

PR #481, opened 2025-11-12
Dtedder/adding unit testing

### LINKED ISSUE #480: Add UnitTests
Add a sub-project for unit tests to the project.

dtedder:
Looking for a suitable approach to incorporating some tests that can be applied to DSA. I wasn't able to find a way to make use of the Test Case explorer in QtCreator. But the application output window functions correctly and the issues tab does have links to tests that fail. This is as close as I could get to a proper test experience without a big change to the application structure so it would be possible to use CMake for the test project where a target can be declared for each test case.

### DIFF
--- a/UnitTests/UnitTests.pro
+++ b/UnitTests/UnitTests.pro
@@ -18,10 +18,7 @@ TEMPLATE = app
 QT += core gui opengl xml testlib network positioning sensors multimedia quick qml xml concurrent
 CONFIG += c++17
 
-ARCGIS_RUNTIME_VERSION = 200.8.0
-DEFINES += ARCGIS_MAPS_SDK_VERSION=$$ARCGIS_RUNTIME_VERSION
 include($$PWD/../Shared/build/arcgisruntime.pri)
-
 include($$PWD/../3rdparty/zlib-ng/zlib_minizip_ng.pri)
 
 INCLUDEPATH += \
--- a/UnitTests/UnitTests.pro
+++ b/UnitTests/UnitTests.pro
@@ -1,0 +1,59 @@
+# Copyright 2012-2025 Esri
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+TARGET = DSA_UnitTests
+TEMPLATE = app
+
+QT += core gui opengl xml testlib network positioning sensors multimedia quick qml xml concurrent
+CONFIG += c++17
+
+ARCGIS_RUNTIME_VERSION = 200.8.0
+DEFINES += ARCGIS_MAPS_SDK_VERSION=$$ARCGIS_RUNTIME_VERSION
+include($$PWD/../Shared/build/arcgisruntime.pri)
+
+include($$PWD/../3rdparty/zlib-ng/zlib_minizip_ng.pri)
+
+INCLUDEPATH += \
+    $$PWD/3rdparty \
+    $$PWD/../Shared \
+    $$PWD/../Shared/alerts \
+    $$PWD/../Shared/analysis \
+    $$PWD/../Shared/messages \
+    $$PWD/../Shared/packages \
+    $$PWD/../Shared/utilities \
+    $$PWD/../Shared/markup \
+    $$PWD/../Shared/configurations
+
+HEADERS += \
+    $$files($$PWD/test_*.h) \
+    $$files($$PWD/../Shared/*.h) \
+    $$files($$PWD/../Shared/alerts/*.h) \
+    $$files($$PWD/../Shared/analysis/*.h) \
+    $$files($$PWD/../Shared/messages/*.h) \
+    $$files($$PWD/../Shared/packages/*.h) \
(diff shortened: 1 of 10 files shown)


UnitTests/UnitTests.pro
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+TARGET = DSA_UnitTests
+TEMPLATE = app
+
+QT += core gui opengl xml testlib network positioning sensors multimedia quick qml xml concurrent
+CONFIG += c++17
+
+ARCGIS_RUNTIME_VERSION = 200.8.0
```
JamesMBallard: Something to think about, but doesn't need to happen now. Consider centralizing this at some point.

UnitTests/test_Message.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <QString>
+#include <QTest>
+
+using namespace Dsa;
+namespace esri_rt = Esri::ArcGISRuntime;
+
+void test_Message::initTestCase()
+{
+
+}
```
JamesMBallard: For brevity you can consider leaving out empty `initTestCase` and `cleanupTestCase` methods. Once you have setup or teardown code to add, you can add them back.

UnitTests/test_Message.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+void test_Message::test_constructor() const
+{
+  // Message()
+  Message mEmpty{};
```
JamesMBallard:
```suggestion
  const Message mEmpty{};
```

UnitTests/test_Message.cpp
```diff
@@ -0,0 +1,139 @@
...
+  QCOMPARE(mEmpty.geometry().isEmpty(), true);
+  QCOMPARE(mEmpty.isEmpty(), true);
+  QCOMPARE(mEmpty.messageAction(), Message::MessageAction::Unknown);
+  QCOMPARE(mEmpty.messageId().isEmpty(), true);
+  QCOMPARE(mEmpty.messageName().isEmpty(), true);
+  QCOMPARE(mEmpty.messageType().isEmpty(), true);
+  QCOMPARE(mEmpty.symbolId().isEmpty(), true);
+
+  // Message(MessageAction messageAction, const Esri::ArcGISRuntime::Geometry& geometry);
+  Message mActionGeometry{Message::MessageAction::Remove, esri_rt::Point{0, 0}};
```
JamesMBallard:
```suggestion
  const Message mActionGeometry{Message::MessageAction::Remove, esri_rt::Point{0, 0}};
```

UnitTests/test_Message.cpp
```diff
@@ -0,0 +1,139 @@
...
+  QCOMPARE(mActionGeometry.messageAction(), Message::MessageAction::Remove);
+  QCOMPARE(mActionGeometry.messageId().isEmpty(), true);
+  QCOMPARE(mActionGeometry.messageName().isEmpty(), true);
+  QCOMPARE(mActionGeometry.messageType().isEmpty(), true);
+  QCOMPARE(mActionGeometry.symbolId().isEmpty(), true);
+}
+
+void test_Message::test_isEmpty() const
+{
+  Message m{Message::MessageAction::Remove, esri_rt::Point{}};
```
JamesMBallard: Make everything const throughout if you can.

UnitTests/test_Message.h
```diff
@@ -0,0 +1,48 @@
...
+// Qt headers
+#include <QObject>
+
+class test_Message : public QObject
+{
+  Q_OBJECT
+
+  inline static const QString TEST_ATTRIBUTE_NAME_1 = QStringLiteral("TEST_ATTRIBUTE_NAME_1");
+  inline static const QString TEST_ATTRIBUTE_VALUE_1 = QStringLiteral("TEST_ATTRIBUTE_VALUE_1");
+  inline static const QString TEST_STRING_1 = QStringLiteral("TEST_STRING_1");
```
JamesMBallard: Consider moving these to the .cpp file if you can. Not a big deal.

UnitTests/test_Configuration.cpp
```diff
@@ -0,0 +1,111 @@
...
+using namespace Dsa;
+
+void test_Configuration::initTestCase()
+{
+}
+
+void test_Configuration::test_constructor()
+{
+  //Configuration(const QString&, const QString&, bool, bool, int);
+  Configuration c{TEST_CONFIGURATION_NAME_1, TEST_CONFIGURATION_URL_1, true, true, 100};
```
JamesMBallard: Optional, but typically what we do in tests is store constants, that way there is less hard-coding values and less chances for mistakes. Probably overkill, but a good practice.

For example:
```c++
constexpr auto selected = true;
Configuration c{TEST_CONFIGURATION_NAME_1, TEST_CONFIGURATION_URL_1, selected, true, 100};
// ...
QCOMPARE(c.selected(), selected);
```

And I'm serious about this being optional. If it's not useful, it just feels like overengineering things.

JamesMBallard (approved):
A few comments, but all optional stuff.

UnitTests/test_Message.h
dtedder: Yes. These don't serve the same purpose as normal class constants. They don't need to be in the header. Moving to .cpp. I only had them here out of habit.

UnitTests/test_Configuration.cpp
dtedder: Maybe just keep to the QStrings and specific numbers for now. But will keep this in mind as we add more cases where it might make sense to have a grouping of values for verifying in QCOMPARE.

UnitTests/UnitTests.pro
dtedder: I think the `Shared/build/arcgisruntime.pri` is probably where this belonged to begin with. I added a commit to move it there.

PR #486, opened 2026-01-08
Dtedder/455 use calcite font

### LINKED ISSUE #455: Update icons
See if we can leverage Calcite icons for DSA, which should give us a cleaner look and feel, and has a lot more icons to choose from as we add more functionality.

We should also look at updating the app overall to use the Calcite styling... using the Theme approach (i.e. if it just works). 

Note there is Calcite styling for Qt toolkit UI controls, but not the Calcite icons, so we need to manually get the icons from the Calcite repo. 

Calcite Icons: https://developers.arcgis.com/calcite-design-system/icons/

Definition of Done:
- [x] Determine the process to get the icons (SVGs) and use them 
- [x] Determine which icons to replace existing ones
- [ ] (TBC) Update overall styling to use Calcite
- [x] Implementation
- [ ] Test
- [ ] Determine if we should keep the new icons
- [ ] Doc

dtedder:
THIS IS AN ARCHIVE PR ONLY. DO NOT MERGE.


dtedder:
work in this branch was done for issue #455. closed without merging. we can revisit this approach at a later time or with a new design.

PR #487, opened 2026-01-15
Adding SSL check for Android

### LINKED ISSUE #439: Better detection for Android builds when SSL libs are missing
Examining the contents of Qt's `openssl.pri` (see below)

<details>
  <summary>openssl.pri</summary>

```qmake
!contains(QT.network_private.enabled_features, openssl-linked) {
    CONFIG(release, debug|release): SSL_PATH = $$PWD
                              else: SSL_PATH = $$PWD/no-asm

    if (versionAtLeast(QT_VERSION, 6.5.0)) {
        ANDROID_EXTRA_LIBS += \
            $$SSL_PATH/ssl_3/arm64-v8a/libcrypto_3.so \
            $$SSL_PATH/ssl_3/arm64-v8a/libssl_3.so \
            $$SSL_PATH/ssl_3/armeabi-v7a/libcrypto_3.so \
            $$SSL_PATH/ssl_3/armeabi-v7a/libssl_3.so \
            $$SSL_PATH/ssl_3/x86/libcrypto_3.so \
            $$SSL_PATH/ssl_3/x86/libssl_3.so \
            $$SSL_PATH/ssl_3/x86_64/libcrypto_3.so \
            $$SSL_PATH/ssl_3/x86_64/libssl_3.so
    } else {
        ANDROID_EXTRA_LIBS += \
            $$SSL_PATH/ssl_1.1/arm64-v8a/libcrypto_1_1.so \
            $$SSL_PATH/ssl_1.1/arm64-v8a/libssl_1_1.so \
            $$SSL_PATH/ssl_1.1/armeabi-v7a/libcrypto_1_1.so \
            $$SSL_PATH/ssl_1.1/armeabi-v7a/libssl_1_1.so \
            $$SSL_PATH/ssl_1.1/x86/libcrypto_1_1.so \
            $$SSL_PATH/ssl_1.1/x86/libssl_1_1.so \
            $$SSL_PATH/ssl_1.1/x86_64/libcrypto_1_1.so \
            $$SSL_PATH/ssl_1.1/x86_64/libssl_1_1.so
    }
}
```

</details>

I think we can add some logic to our DSA .pro files that would examine the `ANDROID_EXTRA_LIBS` var and see if `libcrypto` and `libssl` (wildcarded) are in the list. If they're not, raise an error that SSL is required and wasn't found.

har13205:


Handheld/main.cpp
```diff
@@ -95,6 +95,12 @@ QObject* dsaResourcesProvider(QQmlEngine* engine, QJSEngine* scriptEngine);
 
 int main(int argc, char *argv[])
 {
+#if defined(Q_OS_ANDROID)
+    if (!QSslSocket::supportsSsl()) {
```
dtedder:
```suggestion
    if (!QSslSocket::supportsSsl())
    {
```
A trivial suggestion, but I think we have curly brace on next line pretty much every where except for initializer lists maybe.

Also, the indentation seems one too many. Gets weird when within an ifdef though. So maybe it's fine.

Handheld/main.cpp
```diff
@@ -95,6 +95,12 @@ QObject* dsaResourcesProvider(QQmlEngine* engine, QJSEngine* scriptEngine);
...
+        qCritical() << "OpenSSL libraries not found";
+        return EXIT_FAILURE;
```
dtedder:
```suggestion
        std::abort();
```
Maybe abort makes sense here? I tried this as well as `std:exit(EXIT_FAILURE);` out on a physical device. Didn't get any different behavior though. Just a suggestion. The end result is all probably still the same.

Probably up for debate since it's a pretty ubiquitous header... But I would probably add `#include <cstdlib>` above for any of these options, even if it's just the EXIT_FAILURE macro.

dtedder (approved):

dtedder (changes requested):
minor formatting changes, the rest are just suggestions.

Handheld/main.cpp
har13205: fixed

Handheld/main.cpp
har13205: updated to std::abort();

dtedder: @JamesMBallard `std::abort()` too much here? I'm sure the effective results are more or less the same from an end users perspective. I think `std::abort()` might only be slightly more useful if you are actually using a debugger...

JamesMBallard: I think it's fine to do. We want this error to be obvious and stop users from continuing. Android console logging is notoriously noisy, so messages along are likely to get lost. Either `std::abort` or even `std::terminate()` should be fine.

dtedder (approved):
👍

PR #488, opened 2026-01-28
Dtedder/477 popup updates dynamic entities UI

dtedder:
Changes to the app to rebuild the popup element when a dynamic entity changes


Shared/utilities/common.h
JamesMBallard, line 1: I would remain consistent with filenames, so Common.h. Folders we can be a bit less strict.

Shared/utilities/common.h
```diff
@@ -0,0 +1,29 @@
...
+#include <QPointer>
+#include <QString>
+
+#include <optional>
+#include <tuple>
+#include <variant>
+
+namespace Dsa {
+  namespace ContextMenu {
+    using DynamicEntityPtr = std::optional<QPointer<Esri::ArcGISRuntime::DynamicEntity>>;
```
JamesMBallard: `std::optional<QPointer<...>>` seems redundant. Why do you need to distinguish between there is no value ever, and no value currently? Either way there is a value or not.

Shared/alerts/AlertConditionsController.cpp
```diff
@@ -854,19 +851,19 @@ void AlertConditionsController::onMouseClicked(QMouseEvent &event)
...
     // abort if tool is no longer the active tool
     if (!isActive())
       return;
 
     // iterate over each type in the results variant
-    for (const IdentifyResultsVariant::FutureType& identifyResult : identifyResults)
+    for (const IdentifyResults::Variant& identifyResult : identifyResults)
     {
-      if (identifyResult.index() == IdentifyResultsVariant::Types::LAYERS)
+      if (std::holds_alternative<IdentifyResults::Layer>(identifyResult))
```
JamesMBallard, line 863: I've never seen `std::holds_alternative` before. I had to look it up.

While reading, I found `std::get_if`. Not sure if that would be useful or not.

Shared/analysis/ViewshedController.cpp
```diff
@@ -194,21 +192,21 @@ void ViewshedController::onMouseClicked(QMouseEvent& event)
...
       {
         // bail out if the tool became deactivated
         if (!isActive() || !m_sceneView)
           return;
 
-        for (const IdentifyResultsVariant::FutureType& identifyResult : identifyResults)
+        for (const IdentifyResults::Variant& identifyResult : identifyResults)
         {
-          if (identifyResult.index() == IdentifyResultsVariant::Types::LAYERS)
+          if (std::holds_alternative<IdentifyResults::Layer>(identifyResult))
```
JamesMBallard, line 203: The more I see, I do think std::get_if might be helpful. At a minimum, it may reduce one layer of indenting, because it can combine the type check by returning null if it doesn't match.

Take it with a grain of salt, because I've not used the API before. It could be clunky.

Shared/ContextMenuController.cpp
```diff
@@ -125,9 +124,16 @@ void ContextMenuController::onMousePressedAndHeld(QMouseEvent& event)
     return;
 
   clearOptions();
-  for (const auto& geoElements : std::as_const(m_contextGeoElements))
-    qDeleteAll(geoElements);
-  m_contextGeoElements.clear();
+  std::for_each(std::cbegin(m_contextElements), std::cend(m_contextElements), [&](const ContextMenu::Element& ce)
+  {
+    const ContextMenu::DynamicEntityPtr dePtr = std::get<ContextMenu::DynamicEntityPtr>(ce);
+    if (dePtr.has_value())
```
JamesMBallard: A few thoughts here. As a newcomer to this code, I see 'DynamicEntityPtr' and assume this is some kind of pointer type, but then `dePtr.has_value()` happens so now I need to think about this and see what's really going on.

I do think it's worthwhile to consider dropping std::optional because QPointer can convey "there's no value" just fine in this case.

Shared/ContextMenuController.cpp
```diff
@@ -308,23 +307,44 @@ void ContextMenuController::invokeIdentifyOnGeoView()
           if (geoElements.isEmpty())
             continue;
 
-          // set the GeoElements to be managed by the tool
-          GeoElementUtils::setParent(geoElements, this);
-
-          // add the geoElements to the context hash using the layer name as the key
-          m_contextGeoElements.insert(result->layerContent()->name(), geoElements);
+          // IF/WHEN WE ARE DISPLAYING TRACKS ON THE MAP!!!
```
JamesMBallard: I would reword this.

Shared/ContextMenuController.cpp
```diff
@@ -308,23 +307,44 @@ void ContextMenuController::invokeIdentifyOnGeoView()
...
+              const quint64 eId = de->entityId();
+              const auto itEnd = std::cend(entityIds);
+              if (std::find_if(std::cbegin(entityIds), itEnd, [&](quint64 i) { return eId == i; }) != itEnd)
+                return;
+
+              dePtr = QPointer{deo->dynamicEntity()};
+              geCandidate = de;
+            }
+            else
+              GeoElementUtils::setParent(ge, this);
```
JamesMBallard, line 338:
```suggestion
            else
            {
              GeoElementUtils::setParent(ge, this);
            }
```

Shared/ContextMenuController.cpp
```diff
@@ -308,23 +307,44 @@ void ContextMenuController::invokeIdentifyOnGeoView()
...
+            GeoElement* geCandidate = ge;
+            if (DynamicEntityObservation* deo = dynamic_cast<DynamicEntityObservation*>(ge); deo)
+            {
+              DynamicEntity* de = deo->dynamicEntity();
+              const quint64 eId = de->entityId();
+              const auto itEnd = std::cend(entityIds);
+              if (std::find_if(std::cbegin(entityIds), itEnd, [&](quint64 i) { return eId == i; }) != itEnd)
+                return;
+
+              dePtr = QPointer{deo->dynamicEntity()};
```
JamesMBallard: I would like to think about removing the need for std::optional entirely. If you do end up keeping it, then I would use a make_optional here to make it clearer that's happening on this line.

Shared/ContextMenuController.cpp
```diff
@@ -398,7 +421,11 @@ void ContextMenuController::selectOption(const QString& option)
     if (!identifyTool)
       return;
 
-    identifyTool->showPopups(m_contextGeoElements);
+    // send geoelements to the identify controller which will take ownership of non-dynamic entities
+    identifyTool->setGeoElements(m_contextElements);
```
JamesMBallard, line 433: I don't know if there's any way to avoid this, but it sounds complicated that the identify controller takes ownership of some things and not others.

UnitTests/test_GeoElementUtils.cpp
```diff
@@ -0,0 +1,54 @@
...
+namespace esri_rt = Esri::ArcGISRuntime;
+
+QList<esri_rt::GeoElement*> getListOfGraphics(qsizetype count)
+{
+  QList<esri_rt::GeoElement*> geoElements{};
+  geoElements.reserve(count);
+  for (int i = 0; i < count; ++i)
+    geoElements.emplaceBack(new esri_rt::Graphic());
+  return geoElements;
+}
```
JamesMBallard, line 38: Our typical pattern is to put file local methods into an anonymous namespace. That way if you have another `getListOfGraphics` method in the codebase, they will not conflict with each other, and also you don't need to worry about which one might be called.

UnitTests/test_IdentifyController.cpp
```diff
@@ -0,0 +1,77 @@
...
+static const QString TEST_ATTRIBUTE_VALUE_1 = QStringLiteral("TEST_ATTRIBUTE_VALUE_1");
+
+static bool popupChangedEmitted = false;
+
+bool checkReturnAndReset()
+{
+  bool checkedValue = popupChangedEmitted;
+  popupChangedEmitted = false;
+  return checkedValue;
+}
```
JamesMBallard, line 48: Anonymous namespace

UnitTests/test_MessageFeed.cpp
```diff
@@ -0,0 +1,53 @@
...
+void runFeedMessageTypeTest(const QString& feedName, const QString& feedMessageType, QObject* o)
+{
+  MessageFeed mf{feedName, feedMessageType};
+  o->connect(&mf, &esri_rt::DynamicEntityDataSource::doneLoading, o, [&](const esri_rt::Error&)
+  {
+    QCOMPARE(mf.feedName(), feedName);
+    QCOMPARE(mf.feedMessageType(), feedMessageType);
+  });
+  mf.load();
+  while (mf.loadStatus() == esri_rt::LoadStatus::Loading);
```
JamesMBallard: There are more elegant ways to wait for this. We have a bunch in our internal test suites. See if you can use `QEventLoop` or something similar so it won't busy-wait like this.

Shared/utilities/common.h
dtedder: The optional element of the tuple is a convenience for elsewhere in the code to avoid having to do another dynamic_cast to check for DynamicElement types. So essentially it's a way to distinguish between DynamicElement and all other GeoElement types. One use is when elements need to be reset on longPressHeld. DynamicElement _elements_ don't need to be released but any other types do.

dtedder: Removed based on our discussion.

Shared/alerts/AlertConditionsController.cpp
dtedder: Although `std::get_if` looks much cleaner, the implementation of it is a bit more confusing to look at than `std::holds_alternative`. Sticking with this for now.

Shared/ContextMenuController.cpp
dtedder: Taken care of with the removal of std::optional.

Shared/ContextMenuController.cpp
dtedder: Removed.

UnitTests/test_GeoElementUtils.cpp
dtedder: Added namespace guards.

UnitTests/test_IdentifyController.cpp
dtedder: Added namespace guards

UnitTests/test_MessageFeed.cpp
dtedder: Removed the test as it was not actually testing as intended. Will revisit testing that depends on async.

dtedder:
Several commits added to address the issues mentioned here. But for now we will shelve this PR.

PR #494, opened 2026-03-31
Fix duplicate fields in MessageFeeds

### LINKED ISSUE #493: Duplicate columns in MessageFeeds causing crash on query
The schema definition section of the MessageFeeds contains a bug where a column is added multiple times for the same field name. This causes problems when performing any queries that make use of a where clause.

dtedder:
remove the duplicate fields in the message feed DynamicEntityDataSourceInfo construction

### DIFF
--- a/Shared/messages/MessageFeed.cpp
+++ b/Shared/messages/MessageFeed.cpp
@@ -73,9 +73,8 @@ QFuture<DynamicEntityDataSourceInfo*> MessageFeed::onLoadAsync()
     entity_id_field_name = Message::COT_UID_NAME;
 
     // set the fields
-    field_names.reserve(8);
+    field_names.reserve(7);
     field_names.emplace_back(Message::COT_TYPE_NAME);
-    field_names.emplace_back(Message::COT_UID_NAME);
     field_names.emplace_back(Message::COT_POINT_NAME);
     field_names.emplace_back(Message::COT_POINT_LAT_NAME);
     field_names.emplace_back(Message::COT_POINT_LON_NAME);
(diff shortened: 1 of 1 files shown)


dtedder:
Really nothing that can be verified here at the moment. This doesn't actually cause a problem until trying to invoke a query against the data source using a where clause. Regardless of when querying ends up getting implemented, it should be fixed.

Shared/messages/MessageFeed.cpp
```diff
@@ -73,9 +73,8 @@ QFuture<DynamicEntityDataSourceInfo*> MessageFeed::onLoadAsync()
     entity_id_field_name = Message::COT_UID_NAME;
 
     // set the fields
-    field_names.reserve(8);
+    field_names.reserve(7);
     field_names.emplace_back(Message::COT_TYPE_NAME);
-    field_names.emplace_back(Message::COT_UID_NAME);
```
JamesMBallard, line 78: This one isn't added twice to `field_names` like `GEOMESSAGE_UNIQUE_DESIGNATION_NAME` is. 

Or is the combo of `field_names` + `entity_id_field_name` the entirety we need to deduplicate?

dtedder: The combination with the entity_id_field_name is what has to be considered.

dtedder: We need to keep a copy of the entity id field name to pass to the DynamicEntityDataSourceInfo ctor later. And there is a section below this loop (L99) originally intended to have all fields that are in common to both message types. It ended up being only the `Message::SIDC_NAME` though. I have some code coming up in a future commit that will clean this up a bit more even. Just trying to do the bare minimum here to get it fixed while minimizing the upcoming conflicts.

JamesMBallard (approved):

PR #499, opened 2026-04-15
remove left mouse button check for mobile targets

### LINKED ISSUE #443: Red pulsing graphic not visible when creating observation report
**Steps to reproduce:**

- Open the observation report tool
- Go to the location page
- Click on From Map. When clicked on the view, a red pulsing graphic doesn't appear.

**Platforms**: Android

**Applications**: Vehicle & Handheld

dtedder:
Using OS macro to remove the left button check if iOS or Android

### DIFF
--- a/Shared/messages/ObservationReportController.cpp
+++ b/Shared/messages/ObservationReportController.cpp
@@ -40,6 +40,7 @@
 // Qt headers
 #include <QDateTime>
 #include <QHostInfo>
+#include <QtSystemDetection>
 #include <QUdpSocket>
 #include <QUuid>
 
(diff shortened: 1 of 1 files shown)


Shared/messages/ObservationReportController.cpp
```diff
@@ -40,6 +40,7 @@
 // Qt headers
 #include <QDateTime>
 #include <QHostInfo>
+#include <QtSystemDetection>
```
JamesMBallard, line 43: Typically the Qt OS macros are defined in QtGlobal, but I learned something new if they're brought it from this header as well.

JamesMBallard (commented):
Is this just an optimization, or was this causing a problem on mobile?

dtedder:
Causing a problem on mobile platforms. All taps were getting ignored in this event handler on mobile because there's no concept of a 'left' button. This might also be a problem in some other places in the app where we make map tap an option input of coordinates or for finding a GeoElement with Identify. But this fix is specifically for the observation report tool.

JamesMBallard (approved):

PR #506, opened 2026-05-20
Dtedder/478 479 track display properties and search

### LINKED ISSUE #478: Show previous observations / track line
Allow the user to set track display properties for message feeds (dynamic entity layers). We will need a simple UI design that allows the user to turn on/off some properties but also uses some defaults so it does not get overly complicated. 

(design needed)

Track Display Properties Design:
- Maximum Observations per track
  - Slider (defaults to 5)
  - 'Show all' (sets to 0)
- ~Maximum Duration per track~ _(Removing MaxDuration until issues fixed in the SDK)_
- Toggle to enable displaying previous observations
  - color picker
- Toggle to enable displaying track lines
  - color picker

Expose TrackProperties for message feeds.
- [x] Figure out how to allow the user to change settings for each feed (turn into a panel like the Overlays list?)
- [x] Set maximum observations per track - specify integer or use slider bar.  Default is 5 (including latest observation)
- [x] (optional) Set maximum duration per track - specify time value and time unit. default is 0 (no time duration limit)
- [x] Show previous observations (toggle them on/off) . Default is off.
- [x] Show track line (toggle on/off). Default is off.
- [x] Determine symbology to use for previous observations (default is dark gray circle).
- [x] Determine symbology to use for track line (default is dark gray line - i think)
- [x] Show (static) popup for previous observation (in context menu)
- [x] Persist track display properties in the config file
- [x] (optional) Determine if there is a way to set the renderer in the configuration file for previous observations and track line (i.e. include the json)

### LINKED ISSUE #479: Search/query for specific track id (zoom to Unit xxxx)
Current workflow: 
- Choose a feed from a drop down list
- Search tracks (from that feed ) by entering in a name. 
- As you type, a list is shown with track IDs that match, and all dynamic entities that match are highlighted 
- Click on one of the results, map will zoom to that observation, and context menu will display on the entity
- From there, you can perform action from the callout

dtedder:
- Creates a unified side panel for all map tools; no more overlapping modal dialogs
- Added tabs for the MessageFeeds panel for list, track properties and search/find
- added find functionality in context menu controller from the Find sub-panel for the selected GeoElement
- added new message feed constants for config properties
- created a new property on DSA resources to use for the domain of colors for the drop-down
- moved createRenderer over to the MessageFeed class
- switched to header only for MessageFeedConstants
- added serialization for message feeds


Shared/messages/MessageFeed.cpp
```diff
@@ -19,58 +19,170 @@
...
+    else if (QFile::exists(QString{QStringLiteral("%1/icons/%2")}.arg(m_resourcePath, m_thumbnail)))
+      setThumbnailUrl(QUrl::fromLocalFile(QString{QStringLiteral("%1/icons/%2")}.arg(m_resourcePath, m_thumbnail)));
+  }
+
+  m_messagesOverlay = new MessagesOverlay(this, m_feedMessageType, this);
+
+  // tuple<JsonName, Member*, Default>
+  using PropBool = std::tuple<QString, bool*, bool>;
+  std::vector<PropBool> boolProperties{
+    // { MESSAGE_FEEDS_VISIBLE, &m_isFeedVisible, true },
```
JamesMBallard: Any reason to keep this?

Shared/messages/MessageFeed.cpp
```diff
@@ -19,58 +19,170 @@
...
+MessageFeed::MessageFeed(const QVariantMap& properties, const QString& resourcePath, QObject* parent):
+  DynamicEntityDataSource(parent), m_resourcePath(resourcePath)
 {
+  using namespace MessageFeedConstants;
+  qint8 requiredProperties = 0;
+  qint8 requiredPropertiesValid = 0;
+
+  // set all the string properties and keep track of any required that are not found
+  // tuple<JsonName, Member*, Required, Default, AllowedValues>
+  using PropString = std::tuple<const QString&, QString*, bool, const QString&, const QStringList&>;
```
JamesMBallard: Perhaps QString& would read better than QString*?

Also QString is [copy on write](https://doc.qt.io/qt-6/qstring.html#:~:text=same%20code%20positions.-,Behind%20the%20scenes,-%2C%20QString%20uses%20implicit). Very cheap to copy if you aren't changing it.

Shared/messages/MessageFeedsController.cpp
```diff
@@ -81,6 +79,53 @@ MessageFeedsController::~MessageFeedsController()
...
+  {
+    QStringList resultEntityIds{};
+    QList<DynamicEntity*> entities = result->iterator().asList(this);
+    m_entityIds.reserve(entities.size());
+    std::for_each(std::cbegin(entities), std::cend(entities), [&](const DynamicEntity* entity)
+    {
+      resultEntityIds.append(entity->attributes()->attributeValue(mf->searchAttributeName()).toString());
+      m_entityIds.emplace_back(entity->attributes()->attributeValue(mf->entityIdAttributeName()).toString());
+    });
+    // result->deleteLater();
```
JamesMBallard: Do we still need this?

Shared/messages/MessageFeed.cpp
```diff
@@ -19,58 +19,170 @@
...
+  TrackDisplayProperties* tdp = m_messagesOverlay->trackDisplayProperties();
+  tdp->setShowPreviousObservations(m_showPreviousObservations);
+  updateSymbolObservations();
+  // TRACKLINE
+  tdp->setShowTrackLine(m_showTrackLine);
+  updateSymbolTrackLine();
+  // LIMITS
+  int maximumObservations = m_maximumObservations == 0 ? INT32_MAX : m_maximumObservations;
+  tdp->setMaximumObservations(maximumObservations);
+  // updateDuration();
```
JamesMBallard: Should this remain?

Shared/messages/MessageFeed.cpp
```diff
@@ -19,58 +19,170 @@
...
+  // initialize the track display renderer properties
+  // OBSERVATIONS
+  TrackDisplayProperties* tdp = m_messagesOverlay->trackDisplayProperties();
+  tdp->setShowPreviousObservations(m_showPreviousObservations);
+  updateSymbolObservations();
+  // TRACKLINE
+  tdp->setShowTrackLine(m_showTrackLine);
+  updateSymbolTrackLine();
+  // LIMITS
+  int maximumObservations = m_maximumObservations == 0 ? INT32_MAX : m_maximumObservations;
```
JamesMBallard: Consider `std::numeric_limits<int>::max()` rather than the macros.

Shared/messages/MessageFeed.cpp
```diff
@@ -96,19 +209,18 @@ QFuture<DynamicEntityDataSourceInfo*> MessageFeed::onLoadAsync()
...
-  m_entityIdAttributeName = entity_id_field_name;
-  field_names.emplace_back(entity_id_field_name);
+  field_names.emplace_back(m_entityIdAttributeName);
   field_names.emplace_back(Message::SIDC_NAME);
 
   QList<Field> fields;
-  fields.reserve(field_names.count());
-  for (auto& fn : field_names)
+  fields.reserve(field_names.size());
+  std::for_each(std::cbegin(field_names), std::cend(field_names), [&](const QString& fn)
```
JamesMBallard: Take this advice with a grain of salt, but I prefer simple range-for loops over `std::for_each`. I know the C++ community mantra is to use the standard library whenever possible, and to avoid hand-rolled algorithms. The spirit of those recommendations are to ensure your intent is clear, and mostly to avoid bugs. When it comes to loops, a range-based for loops won't have any bounds concerns and it reads much simpler IMHO.

Also std::for_each takes more args (more room for human error), and a lambda (more room for human error) where a range-loop doesn't need any of that.

I would say both are equally safe and expressive, but one is a bit harder to read without any benefits. Surely this is something all C++ devs can endlessly debate. I am not interested in that, but my preference leans towards simplicity and readability when possible.

Shared/messages/MessageFeed.cpp
```diff
@@ -272,7 +465,7 @@ bool MessageFeed::addMessage(const Message& message)
   {
   case Message::MessageAction::Remove:
     {
-      auto future = this->deleteEntityAsync(messageId);
+      auto future = deleteEntityAsync(messageId);
```
JamesMBallard: You could consider Q_UNUSED to make it clear that you are purposefully not going to handle the completion event for this.

Shared/messages/MessageFeed.cpp
```diff
@@ -307,6 +500,220 @@ bool MessageFeed::addMessage(const Message& message)
...
+    return;
+
+  m_messagesOverlay->trackDisplayProperties()->setShowPreviousObservations(showPreviousObservations);
+  emit feedChanged();
+}
+
+int MessageFeed::maximumObservations() const
+{
+  int maximumObservations = m_maximumObservations;
+  if (maximumObservations == INT32_MAX)
```
JamesMBallard: numerical limits would be a good option here. I see a few. I won't comment on each.

Shared/messages/MessageFeed.cpp
```diff
@@ -307,6 +500,220 @@ bool MessageFeed::addMessage(const Message& message)
...
+
+  QColor c{m_colorObservations};
+  if (!c.isValid())
+    return;
+
+  SimpleRenderer* renderer = dynamic_cast<SimpleRenderer*>(m_messagesOverlay->trackDisplayProperties()->previousObservationRenderer());
+  std::unique_ptr<SimpleMarkerSymbol> symbol = std::make_unique<SimpleMarkerSymbol>();
+  symbol->setColor(c);
+  symbol->setSize(static_cast<float>(m_sizeObservations));
+  renderer->setSymbol(symbol.get());
```
JamesMBallard: I don't think we should be deleting the `SimpleMarkerSymbol` like this. If the intent is to use it, then its lifetime should match our intent. This will probably work this way, but only because of internal implementation details of how we cache objects. That could change, so best not to rely on that.

I wouldn't want this pattern to become the norm. The lifetime should follow the intent.

Shared/messages/MessageFeed.cpp
```diff
@@ -307,6 +500,220 @@ bool MessageFeed::addMessage(const Message& message)
...
+
+  QColor c{m_colorTrackLine};
+  if (!c.isValid())
+    return;
+
+  SimpleRenderer* renderer = dynamic_cast<SimpleRenderer*>(m_messagesOverlay->trackDisplayProperties()->trackLineRenderer());
+  std::unique_ptr<SimpleLineSymbol> symbol = std::make_unique<SimpleLineSymbol>();
+  symbol->setColor(c);
+  symbol->setWidth(static_cast<float>(m_sizeTrackLine));
+  renderer->setSymbol(symbol.get());
```
JamesMBallard: Ditto RAII pattern with SimpleLineSymbol. I won't comment on each of these, but our usage intent and the object lifetime should match.

Shared/messages/MessageFeed.cpp
```diff
@@ -307,6 +500,220 @@ bool MessageFeed::addMessage(const Message& message)
...
+
+// int MessageFeed::maximumDuration() const
+// {
+//   return m_maximumDuration;
+// }
+
+// void MessageFeed::setMaximumDuration(int duration)
+// {
+//   if (m_maximumDuration == duration)
+//     return;
```
JamesMBallard: Should we keep all this?

Shared/messages/MessageFeed.h
```diff
@@ -41,8 +42,18 @@ class MessageFeed : public Esri::ArcGISRuntime::DynamicEntityDataSource
...
 
+  Q_PROPERTY(bool showPreviousObservations READ showPreviousObservations WRITE setShowPreviousObservations NOTIFY feedChanged)
+  Q_PROPERTY(QString colorObservations READ colorObservations WRITE setColorObservations NOTIFY feedChanged)
+  Q_PROPERTY(int sizeObservations READ sizeObservations WRITE setSizeObservations NOTIFY feedChanged)
+  Q_PROPERTY(bool showTrackLine READ showTrackLine WRITE setShowTrackLine NOTIFY feedChanged)
+  Q_PROPERTY(QString colorTrackLine READ colorTrackLine WRITE setColorTrackLine NOTIFY feedChanged)
+  Q_PROPERTY(int sizeTrackLine READ sizeTrackLine WRITE setSizeTrackLine NOTIFY feedChanged)
+  Q_PROPERTY(int maximumObservations READ maximumObservations WRITE setMaximumObservations NOTIFY feedChanged)
+  // Q_PROPERTY(int maximumDuration READ maximumDuration WRITE setMaximumDuration NOTIFY feedChanged)
+  // Q_PROPERTY(QString maximumDurationUnits READ maximumDurationUnits WRITE setMaximumDurationUnits NOTIFY feedChanged)
```
JamesMBallard: There is a lot of commented out stuff in here. Please make a pass and clean things up.

Shared/messages/MessageFeedConstants.h
```diff
@@ -22,24 +22,52 @@
...
+  inline static const QStringList MESSAGE_FEEDS_TRACK_DISPLAY_COLORS{
+    QStringLiteral("#e41a1c"), // red
+    MESSAGE_FEEDS_TRACK_DISPLAY_COLOR_DEFAULT,
+    QStringLiteral("#4daf4a"), // green
+    QStringLiteral("#984ea3"), // purple
+    QStringLiteral("#ff7f00"), // orange
+    QStringLiteral("#ffff33"), // yellow
+    QStringLiteral("#a65628"), // brown
+    QStringLiteral("#f781bf"), // pink
+    QStringLiteral("#999999"), // gray
```
JamesMBallard, line 58: There are a few options for colors.

https://doc.qt.io/qt-6.8/qcolorconstants.html
https://doc.qt.io/qt-6.8/qt.html#GlobalColor-enum

Maybe some of these would work, but if you need to serialize these then the strings are probably ideal.

Shared/messages/MessageFeedsController.cpp
```diff
@@ -182,38 +232,23 @@ QString MessageFeedsController::toolName() const
...
-    if (!rendererThumbnail.isEmpty())
-    {
-      if (QFile::exists(QString(":/Resources/icons/xhdpi/message/%1").arg(rendererThumbnail)))
-        feed->setThumbnailUrl(QString("qrc:/Resources/icons/xhdpi/message/%1").arg(rendererThumbnail));
-      else if (QFile::exists(m_resourcePath + QString("/icons/%1").arg(rendererThumbnail)))
-        feed->setThumbnailUrl(QUrl::fromLocalFile(m_resourcePath + QString("/icons/%1").arg(rendererThumbnail)));
-      else
-        emit toolErrorOccurred(QString("Failed to find icon %1").arg(rendererThumbnail), QString("Could not find icon %1 for feed %2").arg(rendererThumbnail, feedName));
-    }
+  if (m_messageFeeds->count() > 0)
```
JamesMBallard: Is there an isEmpty?

Shared/messages/MessageFeedsController.cpp
```diff
@@ -385,92 +422,38 @@ void MessageFeedsController::setLocationBroadcastInDistress(bool inDistress)
   emit locationBroadcastInDistressChanged();
 }
 
-SurfacePlacement MessageFeedsController::toSurfacePlacement(const QString& surfacePlacement)
+MessageFeed *MessageFeedsController::selectedFeed()
```
JamesMBallard:
```suggestion
MessageFeed* MessageFeedsController::selectedFeed()
```

Shared/qml/MessageFeeds.qml
```diff
@@ -25,41 +26,372 @@ DsaPanel {
...
+                //     Layout.row: 4
+                //     Layout.column: 1
+                //     currentIndex: model.indexOf(toolController.selectedFeed.maximumDurationUnits)
+                //     onCurrentIndexChanged: {
+                //         if (!toolController.selectedFeed)
+                //             return;
+
+                //         toolController.selectedFeed.maximumDurationUnits = model[currentIndex]
+                //     }
+                // }
```
JamesMBallard: Worth keeping around?

Shared/messages/MessageFeed.cpp
dtedder: removed

Shared/messages/MessageFeed.cpp
dtedder: removed

Shared/messages/MessageFeed.cpp
dtedder: removed

Shared/messages/MessageFeed.h
dtedder: removed

Shared/qml/MessageFeeds.qml
dtedder: removed

Shared/messages/MessageFeedConstants.h
dtedder: I couldn't find any matches in the available Qt options. Some might be close enough. But they do end up getting serialized in the end anyways to preserve the user's selection for track line and observation symbol color.

Shared/messages/MessageFeedsController.cpp
dtedder: Yes. Will switch it out.

Shared/messages/MessageFeedConstants.h
JamesMBallard: Leave it be in that case. Code convenience shouldn't drive the colors we want to use.

Shared/messages/MessageFeedsController.cpp
dtedder: Must have took this out during debugging. It can be deleted safely. The DynamicEntities themselves have their own parent, the MessageFeed that created them.

Shared/messages/MessageFeed.cpp
dtedder: Agreed. & looks much nicer here. Likely every object is going to be modified here when the feed properties are hydrated from the config file values. Same goes for all the other types initialized below. I updated them as well.

Shared/messages/MessageFeed.cpp
dtedder: Fair enough. I don't mind the syntax necessarily. But agree, it's really just more to read without any benefit. I guess I just thought this style was the preferred style these days. I'll swap the ones out in this PR and evaluate each new loop going forward on a case by case basis.

Shared/ContextMenuController.cpp
```diff
@@ -15,6 +15,11 @@
  ******************************************************************************/
 
 // PCH header
+#include "DynamicEntityIterator.h"
+#include "DynamicEntityQueryParameters.h"
+#include "DynamicEntityQueryResult.h"
+#include "GlobeCameraController.h"
+#include "Viewpoint.h"
 #include "pch.hpp"
```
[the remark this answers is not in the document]
dtedder: used suggestion button in Qt Creator to grab newly required headers and never went back to clean them up.

Shared/messages/MessageFeedsController.h
```diff
@@ -30,33 +32,38 @@ class QAbstractListModel;
...
 {
   Q_OBJECT
 
   Q_PROPERTY(QAbstractListModel* messageFeeds READ messageFeeds CONSTANT)
+  Q_PROPERTY(QAbstractListModel* entityIdResults READ entityIdResults CONSTANT)
   Q_PROPERTY(bool locationBroadcastEnabled READ isLocationBroadcastEnabled WRITE setLocationBroadcastEnabled NOTIFY locationBroadcastEnabledChanged)
   Q_PROPERTY(int locationBroadcastFrequency READ locationBroadcastFrequency WRITE setLocationBroadcastFrequency NOTIFY locationBroadcastFrequencyChanged)
   Q_PROPERTY(bool locationBroadcastInDistress READ isLocationBroadcastInDistress WRITE setLocationBroadcastInDistress NOTIFY locationBroadcastInDistressChanged)
+  Q_PROPERTY(int selectedFeedIndex READ selectedFeedIndex WRITE setSelectedFeedIndex NOTIFY selectedFeedChanged)
+  Q_PROPERTY(MessageFeed* selectedFeed READ selectedFeed NOTIFY selectedFeedChanged);
```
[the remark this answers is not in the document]
dtedder: switched to parameterless signal. was not used anyways.

Shared/DsaController.cpp
```diff
@@ -408,75 +408,99 @@ void DsaController::writeDefaultConditions()
  */
 void DsaController::writeDefaultMessageFeeds()
 {
-  m_dsaSettings[MessageFeedConstants::MESSAGE_FEED_UDP_PORTS_PROPERTYNAME] = QStringList { QString("45678"), QString("45679") };
+  using namespace Dsa::MessageFeedConstants;
+  m_dsaSettings[MESSAGE_FEED_UDP_PORTS_PROPERTYNAME] = QVariantList{ 45678, 45679 };
```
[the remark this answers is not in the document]
dtedder: Keeping numeric outputs but switching the reading of the config to ensure that items/variants in the list are convertible to int.

Vehicle/qml/main.qml
```diff
@@ -445,47 +490,19 @@ Vehicle {
...
-                viewshedTool.state = "clear";
-                reportToolRow.state = "clear";
-                markupToolRow.state = "clear";
-                addConfigurationTool.state = "clear";
+                // mapToolRow.state = "clear";
+                // alertToolRow.state = "clear";
+                // viewshedTool.state = "clear";
+                // reportToolRow.state = "clear";
+                // markupToolRow.state = "clear";
+                // addConfigurationTool.state = "clear";
```
JamesMBallard: Did you mean to leave these in?

Shared/messages/MessageFeed.cpp
```diff
@@ -19,58 +19,166 @@
...
+MessageFeed::MessageFeed(const QVariantMap& properties, const QString& resourcePath, QObject* parent):
+  DynamicEntityDataSource(parent), m_resourcePath(resourcePath)
 {
+  using namespace MessageFeedConstants;
+  qint8 requiredProperties = 0;
+  qint8 requiredPropertiesValid = 0;
+
+  // set all the string properties and keep track of any required that are not found
+  // tuple<JsonName, Member*, Required, Default, AllowedValues>
+  using PropString = std::tuple<const QString&, QString&, bool, const QString&, const QStringList&>;
```
JamesMBallard, line 72: My suggestion may have made things a bit scary. I didn't like QString pointers, but QString references which are default initialized with `QString{}` will be problematic. What are your thoughts? If this is only a theoretical concern and every entry will be initialized, then I'm not worried about it.

If it's possible we could end up with dangling references, then we should rethink this.

Shared/messages/MessageFeed.cpp
```diff
@@ -245,9 +364,79 @@ void MessageFeed::setThumbnailUrl(const QUrl& thumbnailUrl)
...
+QString MessageFeed::surfacePlacement() const
+{
+  return m_surfacePlacement;
+}
+
+Renderer* MessageFeed::createRenderer()
+{
+  // hold mil2525 symbol styles as statics to be shared by multiple renderers if needed
+  static DictionarySymbolStyle* dictionarySymbolStyleMil2525c = nullptr;
+  static DictionarySymbolStyle* dictionarySymbolStyleMil2525d = nullptr;
```
JamesMBallard: Consider members for these.

Shared/messages/MessageFeed.cpp
```diff
@@ -307,6 +497,185 @@ bool MessageFeed::addMessage(const Message& message)
...
+
+  SimpleRenderer* renderer = dynamic_cast<SimpleRenderer*>(m_messagesOverlay->trackDisplayProperties()->trackLineRenderer());
+  auto* symbol = new SimpleLineSymbol(this);
+  symbol->setColor(c);
+  symbol->setWidth(static_cast<float>(m_sizeTrackLine));
+  auto* previousSymbol = renderer->symbol();
+  renderer->setSymbol(symbol);
+
+  if (previousSymbol)
+    previousSymbol->deleteLater();
```
JamesMBallard, line 657: This block appears twice. Not sure if that can be reasonably centralized, and if not don't worry about it.

I would recommend something like this, but it is optional.

```suggestion
  if (auto* previousSymbol = renderer->symbol(); previousSymbol)
  {
    previousSymbol->deleteLater();
  }
  renderer->setSymbol(symbol);
```

Shared/messages/MessageFeed.h
```diff
@@ -61,23 +70,68 @@ class MessageFeed : public Esri::ArcGISRuntime::DynamicEntityDataSource
...
+  QString colorTrackLine() const;
+  void setColorTrackLine(const QString& color);
+  int sizeTrackLine() const;
+  void setSizeTrackLine(int symbolSize);
+  void updateSymbolTrackLine();
+  QString searchAttributeName() const;
+  QString entityIdAttributeName() const;
+
+  void setupOverlay();
+  bool configurationWasValid() const;
```
JamesMBallard, line 94: Should all these be public methods? If they're only here to support Q_PROPERTY, then you can make them private.

Vehicle/qml/main.qml
dtedder: Nope. Will remove.

Shared/messages/MessageFeed.cpp
dtedder: I don't think there's any risk of dangling ref here. These non-const refs are really just aliases to the class member being copy-assigned a default, and if the json includes the property it is copy assigned that instead. If this member/config property was also required, we make note of it and compare at the end. If all required properties weren't present, we abort and the feed will not be added to the map.

Shared/messages/MessageFeed.cpp
dtedder: I was avoiding calling deleteLater on the symbol object until after calling setSymbol with the new one. Is that a concern here?

JamesMBallard: No, that's not a concern. deleteLater won't do anything until later anyway. My thought was more about readability. During the review this block jumped out and caught my eye. I see your logic and you can keep it that way if you prefer. I don't feel strongly.

Shared/messages/MessageFeed.h
dtedder: Moved all the ones that could be private to private. The getters are used outside in the json serialization.

Shared/messages/MessageFeed.cpp
dtedder: We want to share these objects across all MessageFeed instances. We could push them outside the function into static _class_ properties. But they aren't really used anywhere but inside this function.

JamesMBallard: Ok, I think official static inline members make sense in that case. Function local statics can occasionally obscure lifetime and hurts readability IMHO. If the class should have some static state, my preference would be to make that official.

Shared/messages/MessageFeed.cpp
dtedder: Will leave as is for now then. 👍

Shared/messages/MessageFeed.cpp
dtedder: Moved to private class members.

JamesMBallard (approved):

PR #513, opened 2026-05-28
Updating temp files directory

### LINKED ISSUE #510: Multiple cache files are being created in the data folder
- Open the application and download the data
- A few cache files will be generated initially.
- As you continue interacting with the application, the number of cache files increases.

Image

har13205:
This change updates basemap thumbnail generation to create temporary TileCache*.png files in the OS temp directory instead of the active DSA data folder.

### DIFF
--- a/Shared/TileCacheListModel.cpp
+++ b/Shared/TileCacheListModel.cpp
@@ -24,6 +24,7 @@
 #include "TileCache.h"
 
 // Qt headers
+#include <QDir>
 #include <QFileInfo>
 #include <QTemporaryFile>
 #include <QUrl>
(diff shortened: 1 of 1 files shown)


Shared/TileCacheListModel.cpp
```diff
@@ -109,7 +110,7 @@ bool TileCacheListModel::append(const QString& pathToTileCache)
     if (image.isNull())
       return;
 
-    QTemporaryFile* tempImgFile = new QTemporaryFile("TileCacheXXXXXX.png", tileCache);
+    QTemporaryFile* tempImgFile = new QTemporaryFile(QDir::temp().filePath("TileCacheXXXXXX.png"), tileCache);
```
dtedder:
```suggestion
    QTemporaryFile* tempImgFile = new QTemporaryFile(QDir::temp().filePath(QStringLiteral("TileCacheXXXXXX.png")), tileCache);
```

dtedder (approved):
Confirmed switching between this branch and `v.next`. The image files are not created in the root directory of the app with this branch.

Shared/TileCacheListModel.cpp
dtedder: Minor suggestion.

PR #540, opened 2026-06-18
update intial location

### LINKED ISSUE #522: Home Option do not go to initial viewpoint after package reset
After opening a scene package and reset to default. When clicked on the Home option, the viewpoint does not go back to default.

https://github.com/user-attachments/assets/4753ae28-d6a5-4626-87f9-ba4135a92f42

har13205:
Updating initial location to default viewpoint while resetting the scene

### DIFF
--- a/Shared/DsaController.cpp
+++ b/Shared/DsaController.cpp
@@ -316,6 +316,8 @@ void DsaController::resetToDefaultScene()
   if (!basemapTool)
     return;
 
+  writeInitialLocation(viewpointFromJson(defaultViewpoint()));
+
   // create scene
   Scene* newScene = new Scene(this);
   if (const auto initialViewpoint = readInitialLocation(); !initialViewpoint.isEmpty())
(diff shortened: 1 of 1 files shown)


JamesMBallard (approved):

Shared/DsaController.cpp
```diff
@@ -597,13 +597,7 @@ void DsaController::updateInitialLocationOnSceneChange(bool isInitialization)
...
   if (!isInitialization)
   {
-    // This is a user-defined scene change, so we take the scene's
-    // position as our new source of truth.
-    auto v = m_scene->initialViewpoint();
-    if (!v.isEmpty())
-    {
-      writeInitialLocation(m_scene->initialViewpoint());
-    }
+    return;
```
dtedder, line 599: Not able to make a proper github suggestion here and not truly required to achieve the required outcome. But some indentation is no longer necessary. Also this value returned by `readInitialLocation()` wasn't being captured/reused.
```
  if (!isInitialization)
    return;

  // This is an initilization scene change, replace the scene's
  // initial location with our current initial location.
  if (const Viewpoint v = readInitialLocation(); !v.isEmpty())
  {
    // Note use of setViewPoint instead of setInitialLocation. The latter
    // only works if the scene is not loaded, but all MSPK scenes are loaded with
    // the MSPK so it can't be used.
    Q_UNUSED(geoView->setViewpointAsync(v, 0));
  }
```

PR #541, opened 2026-06-21
Popup definitions and location broadcast fix

kerryrobinson:
Several updates in this PR for you to consider and clean up. I put comments in the code, but the summary is: 

- Popup definitions. These are updated. Should be good for now, but will be easy to update if we want
- CoT popup and schema updates.  Added the popup definition and code to read it, also updated the schema to include some additional fields
- Change to test creating a popup with a separate popup title for DEs vs DEOs. I want to make sure the DE live updates are obvious from the previous observations.  The popup definition list is indexed by feed type, so they end up using the same title. It's working without using the list, so this fix still needs to be made to differentiate DE and DEO in the index.
- Location broadcast messages - updated so that the schema and values match the latest geomessage spec, so that the popup definition works on them as well as messages from the simulator

### DIFF
--- a/Shared/DsaController.cpp
+++ b/Shared/DsaController.cpp
@@ -492,7 +492,10 @@ void DsaController::writeDefaultMessageFeeds()
   m_dsaSettings[MESSAGE_FEEDS_PROPERTYNAME] = messageFeedsJson;
 
   QJsonObject locationBroadcastJson;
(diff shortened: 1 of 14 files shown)


dtedder:
These changes have been incorporated into the commits here:
f7d944d5fb221a13a0b2feec67bc863fa14dbecb
a3ed6dd89ea90aae86149d989adc0243f108f6e0

PR #550, opened 2026-07-07
Improve selected state visibility for shared tool icons

### LINKED ISSUE #524: Convert XY (and other?) tools underline too hard to see on mobile
Convert XY (and other?) tools underline too hard to see on mobile. One you leave the overlays UI, you cannot tell the tool is underlined anymore.

Image

Image

har13205:
Issue - https://github.com/Esri/dynamic-situational-awareness-qt/issues/524

### DIFF
--- a/Shared/qml/ToolIcon.qml
+++ b/Shared/qml/ToolIcon.qml
@@ -29,6 +29,7 @@ Item {
     property bool selected: false
     signal toolSelected()
     property color labelColor: Material.foreground
+    readonly property color selectedLabelColor: Material.accent
 
     Column {
         anchors.fill: parent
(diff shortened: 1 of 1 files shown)


dtedder (approved):
Tried on all platforms except iOS. Looks good to me.

har13205:
@dtedder I increased the height of underline as suggested by @kerryrobinson

dtedder:
> @dtedder I increased the height of underline as suggested by @kerryrobinson

Honestly can't really tell much of a difference on Android. You can see the bottom half of the selection indicator rectangle being cut-off on Windows/MacOS with Handheld though. I think there is some overlap on the main content section and the top row. Maybe that intersection in the UI can be adjusted or better aligned so it doesn't overlap?
image

har13205:
@dtedder I’ve adjusted the layout, please take a look

dtedder:
Looks good. However I would actually set the scaleFactor multiplier on the `indicatorHeight` to something much lower now. With 5x, it looks a bit off centered vertically in the bar. I used 2x multiplier and it looks good on desktop and Android. Now that it's not being partially hidden, it stands out much better even without the size increase.

har13205:
@dtedder updated `indicatorHeight` to 2x multiplier

dtedder (approved):
👍

PR #551, opened 2026-07-10
Dtedder/534 multiple 2d button crash

### LINKED ISSUE #534: Spamming the 2D view button disoreints camera and crashes app
Clicking the 2D button while the camera animation is still in progress will lead to application crash.

dtedder:
prevent passing invalid points to the rotateAround method used in the 2D button on the navigation cluster.

### DIFF
--- a/Shared/NavigationController.cpp
+++ b/Shared/NavigationController.cpp
@@ -20,11 +20,9 @@
 #include "NavigationController.h"
 
 // C++ API headers
-#include "Basemap.h"
 #include "Camera.h"
 #include "ErrorException.h"
 #include "GlobeCameraController.h"
-#include "MapTypes.h"
 #include "OrbitGeoElementCameraController.h"
 #include "OrbitLocationCameraController.h"
 #include "Scene.h"
--- a/Shared/NavigationController.cpp
+++ b/Shared/NavigationController.cpp
@@ -92,24 +92,25 @@ void NavigationController::updateGeoView()
   if (!m_geoView)
     return;
 
+  // There is currently no implementation for MapQuickView. This variable
+  // will remain false if any new features use something other than
+  // the SceneQuickView control.
   m_is3d = false;
+
   m_sceneView = dynamic_cast<SceneView*>(m_geoView);
   if (m_sceneView)
   {
     m_is3d = true;
 
-    Scene* scene = m_sceneView->arcGISScene();
-    if (!scene)
-    {
-      auto* sceneQV = static_cast<SceneQuickView*>(m_sceneView);
-      connect(sceneQV, &SceneQuickView::viewpointChanged, this, [this]()
-      {
-        center();
-      }, Qt::SingleShotConnection);
-      return;
-    }
-
-    center();
+    /*
+     * Note: This connection is only intended to capture a valid geometry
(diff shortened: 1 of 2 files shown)


Shared/NavigationController.cpp
```diff
@@ -318,7 +363,14 @@ void NavigationController::center()
...
 
-    m_currentCenter = point;
+    // if the center of the screen does not intersect the globe and invalid point can be the result
+    // don't update the class property if this happens, fallback is the previous center point
+    if (!point.isEmpty() && point.isValid())
+      m_currentCenter = point;
+
+    // skip if center has never been initialized to a valid point
+    if (m_currentCenter.isEmpty() || !m_currentCenter.isValid())
+      return;
```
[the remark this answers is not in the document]
dtedder: solving a different way but suggestion was valid

Shared/NavigationController.cpp
```diff
@@ -90,15 +92,24 @@ void NavigationController::updateGeoView()
...
+
+    Scene* scene = m_sceneView->arcGISScene();
+    if (!scene)
+    {
+      auto* sceneQV = static_cast<SceneQuickView*>(m_sceneView);
+      connect(sceneQV, &SceneQuickView::viewpointChanged, this, [this]()
+      {
+        center();
+      }, Qt::SingleShotConnection);
+      return;
```
JamesMBallard: I see the logic behind this, but here is another idea to consider. I don't know if this is better, but if there is no scene yet, we can listen (once, via SingleShotConnection) to `sceneChanged`, and then if the scene is loaded or not we can wait for that to happen and center the sceneview.

A potential problem with the current workflow is that the user can pan/zoom and do all kinds of interactions that change the viewpoint before the scene is added or loaded, and that could fire viewpointChanged before it's ready. Is `viewpointChanged` meant to capture "there is now a scene, and it is now loaded" ?

JamesMBallard (commented):
One potential concern with the timing of the `viewpointChanged` logic. Not sure if it's a real problem or hypothetical.

Shared/NavigationController.cpp
dtedder: True, I am taking a short cut here. But so far as I have tested (no basemap, no packages, no default data/config present on startup, new package loaded/unloaded), whenever this signal is emitted, we end up with a valid location that can be used for `center()`, despite the rest of the loading that may still be in progress on things like basemap, operational layers, etc.

dtedder: > Is viewpointChanged meant to capture "there is now a scene, and it is now loaded" ?

Not exactly. It's just meant to give the `NavigationController` a valid starting value for `m_currentCenter`. If we don't, there is an edge case where the user could immediately zoom to ground level, orient the camera above the horizon, click the `2D` button and still create the crash. This signal reliably gives us a good value on initialization.

JamesMBallard: Ok, I recommend leaving in some comments that this is to prevent an edge case and not a primary workflow.

JamesMBallard (approved):

dtedder:
@JamesMBallard added some notes in this section of the code and removed the scene check. It wasn't actually doing anything.

PR #559, opened 2026-07-25
Dtedder/516 popup schema for feeds

### LINKED ISSUE #516: Update Popup for message feeds
Use some predefined json to hard code a nicer looking popup for the message feeds. 

Right now we have the same simplified schema for all message feeds; however, ultimately, we will support the full schema for each feed, so the popups can contain some more interesting information. 
Eventually we may also choose to store the popup config in the DSA config file, but that is not needed for v3.0. 

Expected Behavior:
- [x] Popups display for message feeds according to their message type (CoT, Position Report, Observation, EOD Report, SITREP, etc)
- [x] Location tracks from other DSA apps use the correct popup
- [x] Popups match the definition in the json (in the Resoruces folder)
- [x] Dynamic Entities (latest observations) show a different title (something like "latest updates") than Observations, so that the user knows to expect attributes changing (eventually we will make this experience much better)

DoD:
- [x] Define the popup json (kerry)
- [x] Implement
- [x] Full test with all functionality having to do with message feeds

### LINKED ISSUE #523: App restarts show unique new user to others, stale locations never purged
In a situation where multiple DSA apps are running on the same network, and they are all broadcasting their location to each other, whenever one app restarts it will appear as a new user to everyone else with the same name. Furthermore, the last know location of the app that restarted remains indefinitely on other's maps until they restart the app.

The main requirement is to generate a unique ID per unique designation value (which is the user ID set in DSA settings).

Enhancement: send a fixed device ID as the track id with the location update.  Perhaps generate the ID based on the user name, so it does not change between sessions. 

Fix: Use the `username` as the message ID and the unique designation.

dtedder:
Centralize as many of the message feed settings and constants into the header MessageFeedConstants and group them a bit more logically.
Implement means to add schema and popup definitions as json resource files for message feeds.
Applied new constant references throughout the app and the config file defaults.


Shared/messages/MessageFeedConstants.h
```diff
@@ -19,29 +19,29 @@
 
 // Qt headers
 #include <QString>
+#include <QUrl>
```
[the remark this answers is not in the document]
dtedder: added includes

Shared/Resources/messages/popupDefinitions/sitrep.json
```diff
@@ -0,0 +1,56 @@
...
+        "text": "<span style=\"font-size:12px;\"><i>Reporting unit:</i></span><br><span style=\"font-size:18px;color:#68C1F9\"><strong>{uniquedesignation}</strong></span><br><span style=\"font-size:12px;\"><i>Received: </i><b>{SYS_TIMESTAMP}</b></span>"
+      },
+      {
+        "type": "fields",
+        "fieldInfos": [
+          {
+            "fieldName": "activity",
+            "label": "Activity",
+          },
+          {
```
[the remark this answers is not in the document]
dtedder: didn't notice json editor in Qt Creator wasn't highlighting this as an error.

Shared/messages/MessageFeed.cpp
```diff
@@ -171,52 +173,84 @@ MessageFeed::MessageFeed(const QVariantMap& properties, const QString& resourceP
...
+    return;
+
+  const QVariantMap properties = doc.object().toVariantMap();
+  m_searchAttributeName = properties["attribute_name_search"].toString();
+  m_entityIdAttributeName = properties["attribute_name_id"].toString();
+
+  const QJsonArray attributes = properties["attributes"].toJsonArray();
+  m_fields.reserve(attributes.size() + 2);
+  for (const QJsonValueConstRef attribute : attributes)
   {
```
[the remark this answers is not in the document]
dtedder: the validation on this section was intentionally left a bit open since the resources they parse are a compiled resource. but adding the object check is simple enough to put at the top of the method.

Shared/messages/MessageFeed.cpp
```diff
@@ -489,7 +517,55 @@ bool MessageFeed::addMessage(const Message& message)
...
+        {'S', QStringLiteral("Sea surface")},
+        {'U', QStringLiteral("Sea subsurface")},
+      };
+      QString battleDimension{"Other"};
+      const QChar bdCode = attributes[MessageFeeds::Fields::CoT::TYPE].toString().at(4);
+      if (bdCodes.contains(bdCode))
+        battleDimension = bdCodes[bdCode];
+
+      attributes[MessageFeeds::Fields::CoT::EVENT_TYPE] = QString{"<b><i>Affiliation: </i></b>%1<nbsp/><br><b><i>Battle Dimension: </i></b>%2"}.arg(affiliation, battleDimension);
+    }
```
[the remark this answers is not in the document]
dtedder: added string length checking

Shared/LocationBroadcast.cpp
```diff
@@ -352,13 +353,29 @@ void LocationBroadcast::broadcastLocation()
 
     m_message = Message(Message::MessageAction::Update, m_location);
     m_message.setMessageId(QUuid::createUuid().toString());
```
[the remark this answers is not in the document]
dtedder: switched to username.
https://github.com/Esri/dynamic-situational-awareness-qt/issues/523

Shared/messages/MessageFeed.cpp
```diff
@@ -437,61 +473,111 @@ Renderer* MessageFeed::createRenderer()
...
+    return false;
+  }
 
-  default:
-    if (m_messagesOverlay == nullptr)
-    {
-      emit errorOccurred(Error("MessagesOverlay not set", additionalErrorMessage, ExtendedErrorType::None));
-      return false;
-    }
+  // TODO: validate attributes? message.attributes() will most likely have more than the DEDS schema
```
JamesMBallard, line 520: Anything else to consider with this comment?

Shared/messages/MessageFeedConstants.h
JamesMBallard, line 1: Something to think about for later. global statics are allocated at startup and incur a cost. QString with QStringLiteral is as close to free as you can get, but unordered_map will need more heap allocations.

I don't think these amount of constants are even a measurable factor on DSA startup times, but if this list grows you could consider a lazy-load strategy if any profiling points to these as possible things to refactor.

Shared/messages/MessageFeedsController.cpp
```diff
@@ -107,6 +108,10 @@ void MessageFeedsController::findEntities(const QString& entityIdText)
     m_entityIdResults->setStringList(resultEntityIds);
     mf->messagesOverlay()->selectDynamicEntities(entities);
     params->deleteLater();
+  }).onFailed(this, [params, mf](const ErrorException& error)
+  {
+    params->deleteLater();
+    qWarning() << "Error querying the MessageFeed [" << mf->feedName() << "]:" << error.error().message();
```
JamesMBallard: Optional, but in a recent release we added the ability to qDebug/qWarning our entire Error object. If you just want the message here that's fine, but you can simply do this as well for added details:


```c++
    qWarning() << "Error querying the MessageFeed [" << mf->feedName() << "]:" << error.error();
```

Shared/IdentifyController.cpp
```diff
@@ -180,6 +229,83 @@ bool IdentifyController::canPrev() const
...
+    auto* pd = new PopupDefinition(this);
+    pd->setElements(popupElements);
+    pd->setDescription(popupInfo["description"].toString());
+    pd->setExpressions(expressionInfos);
+    pd->setFields(fieldInfos);
+    pd->setTitle(popupInfo["title"].toString());
+    m_popupDefinitions[url][group] = pd;
+  }
+
+  return m_popupDefinitions[url][group];
```
JamesMBallard, line 306: Is this guaranteed to be valid at this point?

Shared/messages/MessageFeed.cpp
dtedder: I want to leave this here so I can return to it after looking more into the issues with Popup definitions that contain expressions. I think I have ruled out the possibility that any discrepancy in the schema would cause the issue. But I will likely end up ruling that out again during the next investigation.

Shared/messages/MessageFeedConstants.h
dtedder: Good point. Likely not a problem at the moment but will keep an eye on it since it does indeed contain more than just QStringLiteral now.

Shared/IdentifyController.cpp
dtedder: So far as I can tell, there is no way for this to be invalid. If execution somehow reaches this point and the member unordered_map didn't contain the url key and/or the inner map didn't contain the group key, nullptr would be the result. Which is handled in the calling method. Just ends up with a popup that doesn't have the nice formatting on it.

Shared/messages/MessageFeedsController.cpp
dtedder: Nice. I'll switch to that. No need not to display all the info available.

JamesMBallard (approved):

PR #565, opened 2026-08-08
Add simulator app link and other fixes to README.md

### LINKED ISSUE #417: Include Simulator and example XML files as option to download
We should include a zip file of example simulator resources as an option to download as well as the data zip.  This includes the simulator executable for windows with the example XML CoT/GeoMessage files and batch file to start them all up. 

I've already created the zip file and put it here \\apps-data\data\apps\DSA.  We will need to determine if we add it to the existing DSA Data configuration package or as a separate download. 

I don't think this will impact anything except referencing it from the simulator section of the doc. 

Option:
- put xml files in `SimulationData` folder (data package), with batch file
- new portal item for windows simulator app
- update the doc

kerryrobinson:
Updating readme.md to reference the simulator app in the distribution zip, plus some other small grammar fixes, bibliography link fixes, and developer tip Alert blockquote.

_Update: Historically, we have included the Simulator app with the `DSA_Qt_x.x_Windows.zip` file, so we should continue to do that._

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -23,7 +23,7 @@ DSA is an open-source app designed to help you build your own custom field opera
 
 The DSA apps do not require access to a server. All the data and processing are on the device itself. Display of each local data source may be toggled on or off. Teammates share information such as their own location, observation reports, and markup across the peer-to-peer network. The emphasis is on collaboration with other teammates.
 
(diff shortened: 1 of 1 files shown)


dtedder (approved):
Looks good. I like the addition of the `TIP` sections for emphasis.

PR #570, opened 2026-08-13
Fix release build crash

har13205:


Shared/messages/MessageFeed.cpp
```diff
@@ -91,7 +92,7 @@ MessageFeed::MessageFeed(const QVariantMap& properties, const QString& resourceP
     requiredProperties += propertyRequired ? 1 : 0;
     if (const QString s = properties[propertyName].toString(); !s.isEmpty())
     {
-      const auto v = std::get<const QStringList&>(ps);
+      const auto& v = std::get<const QStringList &>(ps);
```
dtedder:
```suggestion
      const auto& v = std::get<const QStringList&>(ps);
```

dtedder (approved):
most minor of comments. 💯

PR #574, opened 2026-08-26
Dtedder/568 configuration updates 2

### LINKED ISSUE #568: Update app behavior when no configuration data is present
## Description
The "File not Found" error message when opening the app for the first time is confusing to the user, and we need to present a more meaningful message. Looking into this more, we also need to make a few changes in how we handle the situation when there is no existing configuration data on the device. 

## Details
Currently, when the app initializes and cannot find the configuration data, it creates the `DSA` folder and the `DsaConfigurations.json` file, as well as the `Default` folder and the `DsaAppSettings.json` file inside it.  Because there is no data in the `Default folder`, when we read the settings file and try to open one of the layers, we show the "File not found" error message (I had assumed that this error message was referring to the app settings json file not being found). Related to this, we had not included a `DsaSettings.json` file in the default sample data portal item zip file, so that when downloaded, it would add the data but not overwrite the config file.  The app config settings file should live with the data so that we can make updates to it without having to change the app code. 

While we are making changes, we should also On the Configurations tab of the Settings panel, the `Close App` button is still not as intuitive. I suggest we update it to say "Close app to load new config" or similar (or another text box appears with the Close button with this message).  We should also look into the delay on the Configurations panel after tapping the Download icon for a portal item. 

## Acceptance Criteria
- [x] The `DsaAppSettings.json` file is added to the default configuration portal item zip (so that it can be updated together with the data if needed)
- [x] Updated app startup behavior:
  - [x] If the `DSA` data folder does not exist, an error message is displayed to the user: "App configuration not found.  Download the default configuration data from Esri ?" (so we never get to the "File not found" error message shown previously)
  - [x] The Default configuration package contains the DsaAppSettings.json file and can overwrite the one created by the app.
  - [x] When the app is restarted, it'll open the Default configuration
  - [x] If the user is offline, they can cancel downloading the default configuration and load a custom configuration by using the + button and pointing to a zip file on the device.  The default configuration will still be listed, but not available to click on until it is downloaded (this is the current behavior, which is correct)
- [x] On the Configurations tab of the Settings panel, when a user selects a different configuration in the list, it is clear to the user that they must close the app and restart it in order to load the new configuration (in addition to the text shown above the list of configurations). New text box next to the `Close App` button that says "_Close and restart the app to load new configuration_". 
- [x] On the configurations tab of the Settings panel, when a user clicks on the Download button, there is instant feedback that the download process has started and they are not able to click it again until the process is complete. 

Discussed:
- Currently, the `DSA\Default` folder must be created with a placeholder `DsaAppSettings.json` file. We can consider not creating the `DSA\Default` folder until the Default configuration package is downloaded by the user, which would mean the app cannot be used until a configuration is loaded - default or otherwise. 

## DoD
- [x] Implementation
- [x] Updated test steps
- [x] Tested on Windows and Android
- [ ] Updated doc (if needed)

dtedder:
- Consolidate the message displayed for the UI as a single string resource
- Prevent the UI from allowing multiple inadvertent downloads for the same configuration
- Update the itemID for the default configuration data download
- Remove the action to create a default configuration file if one is not found in the configuration directory
- Remove the exception that existed for the Esri default download not needing a config file
- Clean up the UI for the configurations list panel

### DIFF
--- a/Handheld/qml/main.qml
+++ b/Handheld/qml/main.qml
@@ -653,7 +653,15 @@ Handheld {
         onAccepted: showConfigurations(true);
         onRejected: showConfigurations(false);
     }
+
+    property bool promptedForDefaultDownload: false
     function showConfigurations(downloadDefaultData) {
+        // prevents multiple 'Yes' taps.
+        // this dialog should only be shown on the initial startup.
+        if (promptedForDefaultDownload)
+            return;
+
+        promptedForDefaultDownload = true;
         if (downloadDefaultData)
             configurationController.downloadDefaultData();
 
--- a/Handheld/qml/main.qml
+++ b/Handheld/qml/main.qml
@@ -34,6 +34,7 @@ Handheld {
     property real hudRadius: 3 * scaleFactor
     property real hudMargins: 5 * scaleFactor
     property bool configurationsChanged: false
+    property bool promptedForDefaultDownload: false
 
     signal clearDialogAccepted();
     signal closeDialogAccepted();
@@ -654,7 +655,6 @@ Handheld {
         onRejected: showConfigurations(false);
     }
 
-    property bool promptedForDefaultDownload: false
     function showConfigurations(downloadDefaultData) {
         // prevents multiple 'Yes' taps.
         // this dialog should only be shown on the initial startup.
--- a/Handheld/qml/main.qml
+++ b/Handheld/qml/main.qml
@@ -649,7 +649,7 @@ Handheld {
 
     DsaYesNoDialog {
         id: configurationDownloadDialog
-        informativeText: "Download the default configuration data from Esri (~450mb)?"
+        informativeText: DsaResources.DefaultConfigurationDownloadPrompt
         onAccepted: showConfigurations(true);
         onRejected: showConfigurations(false);
     }
--- a/Shared/configurations/ConfigurationController.cpp
+++ b/Shared/configurations/ConfigurationController.cpp
@@ -123,11 +123,6 @@ void ConfigurationController::extractConfigurationDownload(const QString& downlo
     if (updateExtractedConfigurationFile(configurationDirectory))
       return;
 
-    // still exit early if the configuration is the default download from Esri
-    const auto configuration = m_configurationListModel->byName(configurationName);
-    if (configuration.urlStr() == ConfigurationController::DEFAULT_DOWNLOAD_URL)
-      return;
-
     // cleanup unusable files and reset the model item for the UI
     resetConfigurationDeviceStatus(configurationName);
 
@@ -161,9 +156,7 @@ void ConfigurationController::extractConfigurationDownload(const QString& downlo
 
 bool ConfigurationController::updateExtractedConfigurationFile(const QDir& configurationDirectory)
 {
-  // any download other than the default from Esri should have it's own DsaAppConfig.json
-  // file present. if it is not present, the default will be created when the app is
-  // relaunched with this configuration selected.
+  // all downloads are required to contain the DsaAppConfig.json in the expected location
   const auto dsaAppConfigFilePath = configurationDirectory.absoluteFilePath(DsaUtility::FILE_NAME_APP_CONFIG);
   QFile dsaAppConfigFile{dsaAppConfigFilePath};
   if (!dsaAppConfigFile.exists() || !dsaAppConfigFile.open(QIODevice::ReadOnly))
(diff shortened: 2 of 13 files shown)


Shared/configurations/ConfigurationController.cpp
```diff
@@ -322,6 +315,12 @@ void ConfigurationController::download(int index)
   const auto configurationUrlStr = configuration.urlStr();
   const auto configurationName = configuration.name();
 
+  // prevent double taps of the download icon from any potential UI lag
+  if (configuration.downloading())
+    return;
+
+  m_configurationListModel->download(configuration.name());
+
```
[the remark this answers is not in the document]
dtedder: These are actively handled by other methods that alter the state of the configuration. But will keep this in mind if any 'stuck' conditions surface.

Shared/qml/Options.qml
```diff
@@ -411,7 +411,7 @@ Rectangle {
                 anchors {
                     fill: parent
                     margins: 10 * scaleFactor
-                    bottomMargin: 10 * scaleFactor + buttonCloseApp.height
+                    bottomMargin: 10 * scaleFactor + restartPromptArea.height
                 }
```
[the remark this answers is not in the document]
dtedder: No loop reported in the QML debugger. Will keep an eye out for this.

Handheld/qml/main.qml
```diff
@@ -649,11 +649,19 @@ Handheld {
 
     DsaYesNoDialog {
         id: configurationDownloadDialog
-        informativeText: "Download the default configuration data from Esri (~450mb)?"
+        informativeText: DsaResources.DefaultConfigurationDownloadPrompt
         onAccepted: showConfigurations(true);
         onRejected: showConfigurations(false);
     }
+
+    property bool promptedForDefaultDownload: false
```
JamesMBallard: Should this be declared higher up? I'm not sure which component this belongs to.

Shared/configurations/ConfigurationListModel.cpp
```diff
@@ -51,6 +51,18 @@ void ConfigurationListModel::cancel(int index)
...
+void ConfigurationListModel::cancel(const QString& configurationName)
+{
+  for (Configuration& c : m_configurations)
+  {
+    if (c.name() != configurationName)
+      continue;
+
+    c.cancelDownload();
+    return;
+  }
```
JamesMBallard, line 63: Can consider replacing raw loop with std::find_if

Shared/qml/Options.qml
```diff
@@ -488,10 +486,23 @@ Rectangle {
                                 anchors.fill: parent
                                 onClicked: {
                                     configurationController.cancel(index);
+                                    if (timerDebounce.running)
+                                        imageDownload.visible = true;
                                 }
                             }
                         }
 
+                        property bool downloadOnCooldown: false
```
JamesMBallard: I don't know what cooldown means here. Also, should this be moved to the top where the object declaration starts? This may be simply personal preference, but typically in QML code I see all the properties grouped at the top.

JamesMBallard (commented):
Minor stuff, but mostly looks good.

Shared/qml/Options.qml
dtedder: Will move up to the start of the `Component`. Term "cooldown" describes the enforced time lapsed on the button click. I've found it's possible to tap the button quickly enough that it will cause more than one download to start before the QML properties are updated on the model object.

Shared/configurations/ConfigurationListModel.cpp
dtedder: Raw loop use here since only one will ever match and we want to perform the cancel and exit as soon as it's found.

Handheld/qml/main.qml
dtedder: Will move to top.

Shared/configurations/ConfigurationListModel.cpp
JamesMBallard: You can do that with std::find_if as well, but it's a bit more code. Conciseness wins here, and the guidance is typically aimed at replacing index-based raw loops, not range-loops. Not as much to gain here anyway.

JamesMBallard (approved):

Shared/configurations/ConfigurationListModel.cpp
dtedder: Sorry, you are right. I was unnecessarily concerned with execution policy here. It is stated clearly in almost every implementation on cppreference.com: `finds the first...`.  Also not sure why I was concerned with a parameter that I wouldn't have even planned on opting into using... 🤔.

But yes, more code, for not much more payoff. I'll leave it as it is on this one.

PR #577, opened 2026-09-03
Dtedder/568 doc updates

### LINKED ISSUE #568: Update app behavior when no configuration data is present

dtedder:
Updated the images showing the configurations list on the options panel. Also a few minor corrections. Unfortunately, I'm not able to do exact recreations as the Mac update I have now includes the 'Liquid Glass' feature and the previous look/feel for some of the menu bars is no longer possible despite all my attempts.

Please have a look @kerryrobinson @har13205


har13205 (approved):

kerryrobinson (approved):
looks good. dsa-tool-settings.png is also updated in Harini's PR.

PR #584, opened 2026-09-16
V.next merged with main

dtedder:
PR for 3.0.0 to main


kerryrobinson (approved):
👍 🚀

har13205 (approved):

ldanzinger (approved):
spot checked, looks good

Handheld/qml/main.qml
```diff
@@ -386,21 +439,58 @@ Handheld {
...
+            Rectangle {
+                anchors {
+                    left: parent.left
+                    right: parent.right
+                    bottom: parent.bottom
+                    margins: 5
+                }
+
+                Button {
+                    text: "Back"
```
jared-2016, line 471:
```suggestion
                    text: qsTr("Back")
```

Not sure how important this is in the context of the DSA app but `qsTr("...")` allows for Internationalization and localization.
https://doc.qt.io/qt-6/internationalization.html
https://doc.qt.io/qt-6/i18n-source-translation.html#qml

So if it's useful you can go back and add this to all strings in qml pretty easily. Otherwise you can just ignore this comment.

Shared/configurations/ConfigurationController.cpp
```diff
@@ -64,77 +79,362 @@ void ConfigurationController::select(int index)
...
     }
 
-    // reset the properties for download status and use the network manager to download the file
-    m_networkReply = m_networkAccessManager.get(networkRequest);
-    m_downloadInProgress = true;
-    connect(m_networkReply, &QNetworkReply::downloadProgress, this, &ConfigurationController::downloadProgress);
-    connect(m_networkReply, &QNetworkReply::readyRead, this, &ConfigurationController::readyRead);
-    connect(m_networkReply, &QNetworkReply::finished, this, &ConfigurationController::finished);
-    connect(m_networkReply, &QNetworkReply::errorOccurred, this, &ConfigurationController::downloadErrorOccurred);
+    m_configurationListModel->setDataByName(configurationName, bytesReceived * 1.0 / bytesTotal * 1.0 * 100, ConfigurationListModel::PercentDownloaded);
```
jared-2016, line 277: curious, why the `* 1.0`?

jared-2016 (approved):
curiosity question and a suggestion which you can take or leave.

JamesMBallard (approved):

dtedder:
Had to push one more commit to update the itemID of the sample data download url. We weren't able to move the item over to the public account ownership and keep the same itemID.

Shared/configurations/ConfigurationController.cpp
dtedder: Since the 'bytes' inputs to this calculation are all ints, this is just to ensure there is no unintended integer division that could end up with a whole number getting rounded to zero resulting in division by zero. Probably a better way write this more expressively.

Handheld/qml/main.qml
dtedder: I think there are probably several places where we should use this. We probably need to create an issue to audit the QML for some improvements like this.

ldanzinger: Yes, definitely not something we want to go about changing right now. New issue sounds good

kerryrobinson:
> Had to push one more commit to update the itemID of the sample data download url. We weren't able to move the item over to the public account ownership and keep the same itemID.

Thanks @dtedder. That was my bad, I should have checked on that beforehand. Looks good to go.

